PR #213, opened 2014-10-02
wording

isomorphisms:
Just some suggestions on the English-language copy.


pcottle:
still feel pretty strong about `we` versus `you`, closing out till its updated

PR #214, opened 2014-10-02
pointer

isomorphisms:
I feel like when I learned that a branch is "just a pointer", that was a "breakthrough moment" for me. Worth changing? The word "pointer" is shorter than "reference" ... on the other hand it's a jargon word not everyone would understand....


pcottle:
I'm down for this -- at first I thought reference was a better word (sine a lot of git terminology is in refs) but coming from a programming background, pointer might be even better

PR #215, opened 2014-10-03
suggestions for wording

isomorphisms:
what does "two unique parents" mean?

what does "include" mean?

merge master into bugfix or bugfix into master?


pcottle:
same as #213

PR #234, opened 2015-01-08
amend was mispelled in Spanish version of mixed2

mgarciaisaia:
Fixes #233.
Thanks to @maxing80 for catching it :)

Issue #233, opened 2015-01-08
--ammend is wrong in the level mixed2


maxing80:
When you read the the second alert, amend is writed --ammend is in the level mixed2

PR #235, opened 2015-01-09
Fixed three typos

mgarciaisaia:
Found them with `git grep ammends`
Related to #234 and #233


pcottle:
ah yeah, of course! didnt check the `spec` folder... i should move that into `src/` someday!

PR #237, opened 2015-01-11
Fix some translation and punctuation for Traditional Chinese.

qcl:
Fix some translation and punctuation for Traditional Chinese (zh_TW).


pcottle:
looks good to me! thanks a bunch :facepunch:

PR #238, opened 2015-01-13
Fix some translation format problems for Chinese and move the 'Fork me' image to local

mdluo:
1. Add blank spaces between Chinese and English words.
2. Replace “gitHub” with “Github”.
3. For amazonaws.com is unaccessible to some region.


mgarciaisaia:
@pcottle, is there any reason why you are not using those [pure CSS 'fork me' ribbons](http://codepo8.github.io/css-fork-on-github-ribbon/) instead of using an image?

pcottle:
> For amazonaws.com is unaccessible to some region.

Oh dang I had no idea! Yeah lets totally just use a local resource. thanks a bunch :D

@mgarciaisaia -- no particular reason, I just didn't know about the pure CSS versions. If we use that I could actually make the label intl-aware (since I think I can tag html elements with `intl-aware` and they live update). Although I have no translations right now for "Fork me on Github" so *shrug*

Let's get this in for now and maybe switch to pure CSS after

pcottle:
The only downside of this is that the image isn't likely to be cached in the browser (compared to the static aws one) but that's a fine downside compared to having the image available globally

PR #239, opened 2015-01-27
use cdn to load libs to speed up

PeterDaveHello:


pcottle:
My only hesitation is that this will prevent offline development :-/ I often work on LGB (or used to) on caltrain / BART / other areas with no internet, so this somewhat prevents actually using the app in those situations.

Maybe we can have this part of the templating step? If we pass in a parameter during build, we could switch between offline and online resources..

PeterDaveHello:
Agree :+1:

PeterDaveHello:
For gh-pages, I think we can just use cdn.

pcottle:
Hey @PeterDaveHello ! Sorry this took a while to get around to but I cranked out a hacky solution today -- I registered two different grunt tasks that pass in an `isProd` parameter to the index templating, which then alternates between the dependencies

I thought about using the git branch name but that seemed a bit hacky, so I went with this. Thanks for the awesome PR! Lemme know if I can give you credit somehow...

PeterDaveHello:
:+1:

PeterDaveHello:
Will you deploy it to http://pcottle.github.com/learnGitBranching/?demo ?

pcottle:
Yep! should be there already:

```
  <!-- Dependencies -->
  <script src="lib/jquery-ui-1.9.0.custom.min.js"></script>
  <script src="lib/backbone.localStorage-min.js"></script>
  <script src="//ajax.googleapis.com/ajax/libs/jquery/1.8.0/jquery.min.js"></script>
<script src="//cdnjs.cloudflare.com/ajax/libs/underscore.js/1.3.3/underscore-min.js"></script>
<script src="//cdnjs.cloudflare.com/ajax/libs/backbone.js/0.9.2/backbone-min.js"></script>
<script src="//cdnjs.cloudflare.com/ajax/libs/raphael/2.1.0/raphael-min.js"></script>

```

is what im getting

PeterDaveHello:
great!

pcottle:
@PeterDaveHello I got an email saying `ajax.googleapis.com` isnt available for chinese users -- do you know if cloudfare works in china? I removed it in 4df523708b5f874becfbf03a0504802793ede126

PeterDaveHello:
I think it may been blocked, I tested it from several different testing site, and github also been blocked before, if we have problem on google cdn only, transferring to cdnjs may be a choice, cdnjs looks fine in china, https://cdnjs.com/libraries/jquery

test result:
(ok)
http://viewdns.info/chinesefirewall/?domain=ajax.googleapis.com
(not okay)
http://www.websitepulse.com/help/testtools.china-test.html
(not okay)
https://en.greatfire.org/test?url=https%3A//ajax.googleapis.com/ajax/libs/jquery/1.11.2/jquery.min.js

pcottle:
Ah cool! I had no idea there were sites to test CDN loads in other countries. im down to switch to cloudfare's jquery since they seem good here:
http://viewdns.info/chinesefirewall/?domain=https%3A%2F%2Fcdnjs.cloudflare.com%2Fajax%2Flibs%2Fjquery%2F2.1.3%2Fjquery.js

do you want to take a crack at it? should be easier now that the build system is split between prod and dev. and then you can be an official contributor :)

PeterDaveHello:
Sure, I did it in #245

PR #245, opened 2015-02-12
Use minified jQuery on cdn for production env

PeterDaveHello:


pcottle:
thanks a bunch! will push in a bit

PeterDaveHello:
:+1:

PR #246, opened 2015-02-13
Teaching the "-b" option in lesson 2

alexed1:
Adds a modal that pops up in level2 and teaches the "-b" command that's required in order to get a perfect solution on level3. Allows the student who's paying attention to solve level1 in one step, beating the standard,
but doesn't get in the way of a student solving it the traditional way. However, the real argument for it is that you never want to put material on the test that you didn't talk about in class. It just seemed like a lost opportunity to make "-b" a requirement for level 3 perfection but never introduce it to the student.

### DIFF
--- a/src/levels/intro/branching.js
+++ b/src/levels/intro/branching.js
@@ -88,11 +88,20 @@ exports.level = {
               "This will put us on the new branch before committing our changes"
             ],
             "afterMarkdowns": [
-              "There we go! Our changes were recorded on the new branch"
-            ],
+              "There we go! Our changes were recorded on the new branch.",
+              ""
+             ],
             "command": "git checkout newImage; git commit",
             "beforeCommand": "git branch newImage"
           }
+        },
+         {
+            "type": "ModalAlert",
+            "options": {
+                "markdowns": [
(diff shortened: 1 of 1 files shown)


pcottle:
Hrm it is a good point that I never covered `git checkout -b` directly, but it seems like not a ton of content to warrant a new modal. and unfortunately this means most of our translations are out of sync :-/

what do you think about just adding it as a note to the end of the lesson or another slide?

alexed1:
I agree that the modal is a little overbearing. my first thought was simply to add it to the bottom of the existing text on the preceding window. The problem is that the window size seems fairly fixed, and there isn't any autoscroll, so a reader wouldn't know it was there. Is that an issue you've grappled with at all? Perhaps I should experiment with trying to scroll the box down a little at the right time...

pcottle:
Yeah I've kept the alert windows fairly fixed so I dont overload any particular slide; i do believe the git demonstration ones scroll though.

What do you think about putting it on this slide?

```
"Ok! You are all ready to get branching. Once this window closes,",
               "make a new branch named `bugFix` and switch to that branch"
```

I think that's maybe an appropriate place for a little extra bonus tip. and that way if you read the tip, you can actually beat the current solution :)

pcottle:
making this part of the objective dialog:

screen shot 2015-11-30 at 10 38 48 am

Issue #249, opened 2015-02-24
404 on main.56afcf29.css 


Insti:
When trying to access: http://pcottle.github.io/learnGitBranching/

The main stylesheet seems to not be found, causing the whole page to fail to render correctly.

http://pcottle.github.io/learnGitBranching/build/main.56afcf29.css

christothes:
I hit this today also

pcottle:
Ah my bad! I'm in manila right now for work and have really bad internet -- for PR #247 yesterday I had to push to prod and something probably got messed up along the way.

i just pushed again and it seems to be working -- sorry for the interruption :scream:

PR #251, opened 2015-02-24
first link tag is pointing to the wrong endpoint

edhedges:
Changed it to main.css


pcottle:
I just fixed this over in Issue #249 -- sorry about the interruption!

We actually need the production endpoint to hit the `{{styleFile}}` since it is hashed for each build, which means it breaks browser cache and makes sure if we have any critical CSS updates they will correctly get shipped.

So can't accept this PR unfortunately but thanks for putting up the fix!

edhedges:
@pcottle My bad. Thanks for the quick! I have never been to this site, but it was linked to me and just figured I'd attempt at helping. I should have checked out the existing issues though.

PR #253, opened 2015-02-26
Update grabbingOneCommit.js

cvogt:


pcottle:
good catch! sometimes this can be true (separate branches all stacked) but i think commits works a bit better here.

its a narrow difference though so the translations being out of sync probably isnt a huge deal

PR #264, opened 2015-05-04
Russian translation

coyl:
• Translated all the levels, except "remote" section.
• Translated all the system messages


pcottle:
This is awesome @coyl! Could you try rebasing so I can see what the final merge would look like? Since we are changing the locale constant I want to make sure it gets addressed everywhere thats needed

also sorry this project is so hard to translate :-/

PR #269, opened 2015-05-08
Fixes in Russian traslation

coyl:
* Added missing levels names
* Fixed typos


coyl:
Also, are you sure, that it's good idea to change untraslated strings with "translate it please" message? Maybe it would be better to change it with an English strings? And optionally make some little mark with modal window or something like that. Now, for example, in Russian version the "Remotes" section consists only of "translate it please" messages.

pcottle:
> Also, are you sure, that it's good idea to change untraslated strings with "translate it please" message? Maybe it would be better to change it with an English strings? 

Yeah I'm open to feedback on this -- it was a pretty brutal last-minute add just so people realized **why** they were seeing it. I think for certain languages it makes sense to fallback, but I was worried about people not knowing why all of the sudden english popped up.

I agree its more noise now though than benefit, so let me remove it

pcottle:
Boom! Now we show the english name + the untranslated warning, should be a lot better

screen shot 2015-05-08 at 12 56 35 pm

i realize that we fall back to english strings for all the normal keys but the level names / hints have custom code for whatever reason. thanks for the feedback though!

PR #271, opened 2015-05-12
Translation/russian

coyl:
* Another part of fixes
* Added myself to contributors list in readme =)


pcottle:
> Added myself to contributors list in readme =)

my bad! I should have added you way back. thanks for the PR

PR #293, opened 2015-06-09
fix some escaping with react views

pcottle:


pcottle:
screen shot 2015-06-09 at 11 20 35 am
seems to work...

pcottle:
screen shot 2015-06-09 at 11 20 54 am

and that of course

PR #299, opened 2015-06-24
Combine dialogs

pcottle:


pcottle:
tested locally, should be good...

PR #310, opened 2015-10-10
translated detachedHead.js in korean

JuhoKang:
translated detachedHead.js in korean

I am trying to start translating the rest of the application into korean because this beautiful tutorial is not translated completely in korean
please let me know if there is something I have to be aware of.


pcottle:
awesome! Thanks so much -- hope other korean native speakers can learn git easier this way.

PR #317, opened 2015-10-15
fix typos

sPaTRoL:


pcottle:
many thanks sir! youc an do these in bigger grouped PRs if you want

PR #325, opened 2015-11-02
Translated tags.js & describe.js into Korean

JuhoKang:
also fixed indentations you mentioned.

### DIFF
--- a/src/levels/mixed/describe.js
+++ b/src/levels/mixed/describe.js
@@ -11,7 +11,8 @@ exports.level = {
     "pt_BR": "Git Describe",
     "zh_TW": "git describe",
     "zh_CN": "git describe",
-    "ru_RU": "Git describe"
+    "ru_RU": "Git describe",
+	"ko"   : "Git describe(묘사)"
   },
   "hint": {
     "en_US": "Just commit once on bugFix when you're ready to move on",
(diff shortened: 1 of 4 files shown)


src/levels/mixed/describe.js
```diff
@@ -1,4 +1,4 @@
-exports.level = {
+﻿exports.level = {
```
pcottle: Hrm what is the change here? hard to tell but it looks like whitespace?

src/levels/rampup/cherryPick.js
```diff
@@ -28,7 +28,7 @@ exports.level = {
     "zh_CN": "git cherry-pick 后面跟着 commit 的名字",
     "zh_TW": "git cherry-pick 後面要接著 commit 的名稱",
     "ru_RU": "git cherry-pick основывается на именах коммитов!",
-	"ko": "커밋의 이름들로 git cherry-pick 하세요!"
+    "ko"   : "커밋의 이름들로 git cherry-pick 하세요!"
```
pcottle: thanks!

pcottle:
Great! Little weird that the first line of each file shows a change -- maybe its just whitespace? the tests pass though so ill trust you :tada: :+1:

JuhoKang:
I'll find out what the "weird invisible change" is next time I work on it.
Maybe it's because of the text editor I used. (notepad++ portable) 
The indentation problem was caused by this editor so the "weird invisible change" might also be

PR #328, opened 2015-11-07
Fix Tranlation issue from Pt-Br in fetchRebase

mgiuglio:
There is a typing error in fetchRebase.js file that makes Portuguese translation unavailable for hint.


pcottle:
thx!!

PR #330, opened 2015-11-14
Translated clone.js & remoteBranches.js into Korean

JuhoKang:


pcottle:
Awesome! thanks a bunch, ill push the updates to the site

PR #335, opened 2015-11-19
Translated fetch.js & pull.js into Korean

JuhoKang:


pcottle:
thanks a bunch! :tada:

PR #344, opened 2015-12-03
Fix some Chinese translation

weisongye:
* Insert space between ASCII characters and non-ASCII characters.

* Fix some Chinese translation.


pcottle:
wow very thorough!! thanks a bunch :tada:

PR #346, opened 2015-12-10
Translated mergeManyFeatures.js & tracking.js into Korean

JuhoKang:


pcottle:
:+1: :tada:  thanks as always

PR #351, opened 2016-01-14
Fixes in russian translation

demidovakatya:
Mostly grammar, typos, punctuation.


pcottle:
:tada: :+1: thanks for helping out! the russian version of the site has really been on the rise and its all thanks to these awesome PR's

screen shot 2016-01-13 at 4 45 51 pm

demidovakatya:
Yay!

I've had some issues with the markdown and want to ask your advice. When do we use inline code? For example, should it be "`git push`" or "git push"?

pcottle:
Honestly im pretty inconsistent with it, but ideally it should be `git push` if referred to as a command in the middle of a sentence, but probably should be normal english words if being referenced. like "If you perform a git push to the server, the code will be updated. Invoke this by running `git push`"

but not a huge deal :)

demidovakatya:
thanks, I think I've got the idea :)

PR #361, opened 2016-02-24
Fix a typo in french translation

zyurs:


pcottle:
Thanks! I don't want to bust cache for everyone by releasing a new revision, so this will go out next time i push the site :)

PR #380, opened 2016-07-05
Fixed bugs and typos in Ukrainian translation

andriymoroz:


src/levels/mixed/jugglingCommits2.js
```diff
@@ -1,512 +1,512 @@
-exports.level = {
```
pcottle: what happened here? do you  know why git shows the whole file as added/removed?

pcottle:
Totally down to merge this except for the diff on `jugglingCommits2.js` and `tags.js`. any idea what went on there? did you change the newlines or something?

pcottle:
yolo merge since i inspected locally and it looked fine

src/levels/mixed/jugglingCommits2.js
andriymoroz: the line endings were in DOS format (CR+LF)
git even refused to apply its own patch created by format-patch if the patch touches these files

PR #382, opened 2016-07-08
Add ja tracking translation

mizunashi-mana:


pcottle:
Thank you @mizunashi-mana!!

PR #384, opened 2016-07-18
Updates zh_CN translation

timothyqiu:
+ Updates Simplified Chinese translation of _Git Commits_ to match the English version.
+ Improves Simplified Chinese translation by fixing ambiguity and reorganizing some words.


pcottle:
looks great! 🎉  thanks :D

Issue #406, opened 2017-01-31
Terminal column is too height on firefox and firefox-dev-edition


enguerran:
It could come from some CSS flex rules. I could not fix this (but I tried with some `max-height` on `#terminal` and children, with no success at all).

Here is a screenshot from firefox-dev-edition 52.0a2, look how the terminal on the left is too height and no scroll could let me add another git command.

*Window border's buttons of the terminal are not well aligned too.*  (but it is not blocking)

![screenshot](http://i.imgur.com/hnvpGKE.png)

This tool is awesome, very good work! Thanks @pcottle.

pcottle:
Yeah this is tricky, we only support Chrome for the time being unfortunately :-/ would gladly accept any PR's that fix up firefox though

enguerran:
I close this issue as you only support Chromium. And it looks like some people try to help: for instance this fixes a little bit the layout https://github.com/pcottle/learnGitBranching/commit/07964bf1981b04547d73e266d0060da55a59579f. Thanks for your amazing tool by the way. 💓

PR #407, opened 2017-02-11
Fix for #406: scroll for the terminal in Firefox

ser-dev:
Seems like Firefox requires the height property in order to display the scroll.


pcottle:
great!! thanks for the fix 🎉

ser-dev:
Thank You for this awesome tool! 👍

PR #414, opened 2017-03-12
Fix showing text cursor after page load

AlexByte:


pcottle:
Woohoo! 🎉  thanks so much

PR #427, opened 2017-05-05
fix chinese translate

hotleave:
- fixed the "INTL out of sync"
- fix some typo
- make it more official, less buzzwords，something like "碉堡了", "牛鼻啊！" or "哥相信你！"


pcottle:
wow this is a huge amount of work! thanks @hotleave  so much 🎉  will pull this down and push it

PR #464, opened 2018-05-10
Add gif

filipefilardi:
Would be nice to replace the image in readme with a gif, impacting first time visitors.

I added the following gif

demo-gif


filipefilardi:
@pcottle i don't think it expire didn't found any expiring info in my search. 

... otherwise i'll need to manage a lot of my repos ;)

pcottle:
alright! yeah I guess if its uploaded user content lets hope not, but I could imagine there being some max TTL. whereas raw.github will always serve it as long as its in the repo. also easier to update :)

PR #473, opened 2018-07-07
commands.js: add --force alias to the -f option (fix #472)

### LINKED ISSUE #472: `git branch -f` works but `git branch --force` doesn't
At least in level `mixed2`, I was getting `The option "--force" is not supported!` but `git branch -f` worked fine.

waldyrious:
Modeled after the current implementation of `branch -d` / `branch -D`

Note: since I'm not familiar with the code, this PR is an experimental submission to address #472; if it's successful, it will hopefully illuminate a viable path for me or others to submit follow-up PRs adding the remaining missing aliases, for consistency, as discussed in that issue thread.

### DIFF
--- a/src/js/git/commands.js
+++ b/src/js/git/commands.js
@@ -413,6 +413,7 @@ var commandConfig = {
       '-d',
       '-D',
       '-f',
+      '--force',
       '-a',
       '-r',
       '-u',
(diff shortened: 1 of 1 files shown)


pcottle:
Great! Thanks a bunch. Donno when I'll push the site next but this will make the next release.

Sorry its a bit hacky to map one alias to another :P ideally theres some abstraction...

waldyrious:
No problem! The usefulness of the tool greatly balances out any implementation wrinkles :)

PR #475, opened 2018-07-10
commands.js: add --all alias to the -a option

waldyrious:
Addresses #472, follow-up of #473.


pcottle:
pushed the  site!
screenshot 2018-07-10 at 2 50 56 pm

PR #490, opened 2018-10-17
Typo fix

moskdev:


pcottle:
Thanks a bunch! Dont know when I’ll push the site next since I’m remote but it’ll go out with the next deploy

PR #492, opened 2018-10-22
Set html lang and meta http-equiv="content-language" attr on language change

PanAeon:
According to [Google documentation](https://support.google.com/webmasters/answer/182192?hl=en) this doesn't change much, as google won't look at this attributes, but it may be helpfull for less sophisticated search engines.

### DIFF
--- a/src/js/app/index.js
+++ b/src/js/app/index.js
@@ -5,6 +5,7 @@ var React = require('react');
 var assign = require('object-assign');
 var util = require('../util');
 var intl = require('../intl');
+var _ = require('underscore');
 var LocaleStore = require('../stores/LocaleStore');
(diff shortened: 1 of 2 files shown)


src/js/app/index.js
```diff
@@ -5,6 +5,7 @@ var React = require('react');
 var assign = require('object-assign');
 var util = require('../util');
 var intl = require('../intl');
+var _ = require('underscore');
```
pcottle: Why do we need the underscore import?

PanAeon:
Yes, you are right about `.split("_")`. I wasn't sure what it would do when the string is empty.

pcottle:
🎉 looks great now!

PR #493, opened 2018-10-22
Generate alternate url for different locales so search engine could find them all.

PanAeon:
According to [google documents](https://support.google.com/webmasters/answer/189077)  a site can notify google about alternative locales, using html files of the form:
`<link rel="alternate" hreflang="lang_code"... >` 

This pull request generates such links for each of the locales defined in the LocaleStore and inserts them to the top of <head> element where search engines can find it. (I hope google is smart enough to parse head again after js loads, otherwise this is kind of in vane.)

### DIFF
--- a/src/js/app/index.js
+++ b/src/js/app/index.js
@@ -1,6 +1,7 @@
 var Backbone = require('backbone');
 var EventEmitter = require('events').EventEmitter;
 var React = require('react');
+var _ = require('underscore');
 
 var assign = require('object-assign');
 var util = require('../util');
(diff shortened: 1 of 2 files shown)


src/js/stores/LocaleStore.js
```diff
@@ -30,6 +31,8 @@ var headerLocaleMap = {
   'pt-BR': 'pt_BR'
 };
 
+var supportedLocalesList = _.values(langLocaleMap).concat(_.values(headerLocaleMap));
```
pcottle: there are some duplicates here -- for instance `zh_CN` -- if we just add the values of each. can you convert to a set? or perhaps just make them all keys in an object?

also we dont need the underscore here since i polyfill

```
var allLocales = {};
Object.values(langLocaleMap).concat(Object.values(headerLocaleMap)).map(locale => allLocales[locale] = true);

// later
return Object.keys(allLocales)
```

src/js/stores/LocaleStore.js
```diff
@@ -3,6 +3,7 @@
 var AppConstants = require('../constants/AppConstants');
 var AppDispatcher = require('../dispatcher/AppDispatcher');
 var EventEmitter = require('events').EventEmitter;
+var _ = require('underscore');
```
pcottle: rm

src/js/app/index.js
```diff
@@ -86,6 +87,22 @@ var vcsModeRefresh = function(eventData) {
...
 
+var insertAlternateLinks = function(pageId) {
+  // For now pageId is null, which would link to the main page.
+  // In future if pageId is provided this method should link to a specific page
+
+  // The value of the hreflang attribute identifies the language (in ISO 639-1 format)
+  // and optionally a region (in ISO 3166-1 Alpha 2 format) of an alternate URL
+  var altLinks = _.map(LocaleStore.getSupportedLocales(), function(langCode) {
+    var url = "https://learngitbranching.js.org/?locale=" + langCode;
+    return '<link rel="alternate" hreflang="'+langCode+'" href="' + url +'" />';
```
pcottle: a lot of this will break if the langCode has a `"` for some reason. maybe we can uri encode or use jquery to set the attributes? will be cleaner than doing string concats

PanAeon: Well, this code should work only with supported locales, so no way we are going to have `"` in the locales. I can escape the langCodes of course, but it looks like it is a bit of an overkill.

PanAeon:
And regarding unerscore - do you mind if I create another pull request to remove the underscore library completely from the code?

pcottle:
sure thatd be great!

PR #499, opened 2018-11-25
decrease steps of solution

AgranatMarkit:
I find more compact solution.


AgranatMarkit:
The initial solution was:
1) Checkout master.
2) Pull.
3) Merge master with side1.
My solution is:
1) Rebase master on side1.
2) Pull.
Net step is identical.

pcottle:
Totally get that this is more efficient and thanks for the PR! However as the level intro highlights, we want to really have a merge-driven solution to this level rather than one that involves any rebasing. its supposed to contrast a merge vs rebase flow, hence I went with rebase commands. lemme know if that explanation doesnt make sense!

PR #503, opened 2018-11-28
update zh_cn translation in describe.js

RogerLiNing:
The original translation missed this "when you're ready to move on", so I want to add this to the simplified Chinese translation.


pcottle:
🎉  thanks a bunch!

PR #507, opened 2018-11-30
Remove module object-assign

Hongarc:
We can use `Object.assign` with Node.js 4 or higher.


pcottle:
great call! yeah this is more straightforward.

PR #509, opened 2018-12-01
Use simple method of Nodejs instead of `underscore`

Hongarc:
I use method of Node.js instead of :

- `_.extend` 
- `_.clone`
- `_.each`
- `_.forEach`
- `_.reduce`
- `_.map`
- `_.filter`
- `_.keys`
- `_.escape`

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -62,7 +62,7 @@ module.exports = function(grunt) {
       hashedMinFile = 'bundle.js';
     }
     var jsRegex = /bundle\.min\.\w+\.js/;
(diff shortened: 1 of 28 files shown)


Hongarc:
I am trying remove `_` in this project, but I don't know about some function of `_`.

z6102075:
是啊，为什么

Hongarc:
@z6102075 , because it will quick and sort than use `underscore`.
I don't hate  `underscore`
`_` is good with old version of Node.js and browser (it don't support that method).
and we just use some function of underscore`.

I am just try, not sure.

PR #510, opened 2018-12-01
Fix space-group in <vcs> <command>

### LINKED ISSUE #506: Wrong args
I type: (2 space in `git` and `checkout`)
```bash
    git  checkout myBranch
````
but it return 2 args of checkout command: `checkout` and `myBranch` (args must be `myBranch`) so `terminal` log:

> I expected at most 2 argument(s) for with git checkout

We need remove git command (`checkout `, `commit`, etc) from args when type many space.

hong4rc:
Fixed #506  when I have many space between <vcs> and <command>

We don't need `join` option and split it again.

### DIFF
--- a/src/js/commands/index.js
+++ b/src/js/commands/index.js
@@ -127,7 +127,7 @@ var parse = function(str) {
         // before our "stuff" we can simply
         // split on space-groups and grab everything after
         // the second:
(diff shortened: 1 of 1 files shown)


Hongarc:
I think that TODO comment for `CommandOptionParser.prototype.explodeAndSet`

PR #511, opened 2018-12-02
Fixed some css wrong

### LINKED ISSUE #478: I'd like to scroll the console.
When it reaches to the bottom of the page, I won't be able to see below anymore. If you could scroll, you'd see below.

Hongarc:
Fix #478.
- When we scale browser to 125%
- Fixed  `#terminal`  will overlay `commandLineBar` if `#terminal` is long.
- Avoid display 900px in small screen


pcottle:
as always, appreciate all these great PRs! 🎉

PR #514, opened 2018-12-02
remove forgot in 'no more native' a9dd27c

Hongarc:
- You forgot remove it in a9dd27c.
- `Gruntfile.js`: no file match that pattern

I created pull request #512, but it fail inn CI because something about version.
You can test it in your local, if it pass I will open it again.


Gruntfile.js
```diff
@@ -109,14 +109,9 @@ module.exports = function(grunt) {
         'Gruntfile.js',
         '__tests__/*.spec.js',
         'src/js/**/*.js',
-        'src/js/**/**/*.js',
-        'src/levels/**/*.js'
```
pcottle: still think hinting the level files is nice, although i guess they are just mostly json

Hongarc: Oh, sorry, I think it is `src,js/levels`, but it is `src/levels`, I will fixed it.

Issue #161, opened 2014-02-11
Make this project more easy to translate


luzfcb:
Thanks for this project. I'm interested in translating it to pt_BR. I have little knowledge of Javascript. I'm a Django (python language) developer. In Django have a simple way of internationalizing applications. ( https://docs.djangoproject.com/en/1.6/topics/i18n/translation/ )

When I want to support the translation of certain text, instead of using a normal string, use a function call to "gettext" passing as parameter the string to be translated. (the example of the link above is using an alias for unicode gettext, in other words _ (" string ")   )

After this step, execute a command for Django to generate the translation files ( .po files) and send to transifex.com translation platform, which is free for opensource projects.
The Transifex platform is very easy to manage translations and groups of translators.

Is it possible to add support for something like this project?
It would be interesting at least list somewhere, the files containing the strings for translation to facilitate the addition of new translations.

What is your opinion about it?

http://stackoverflow.com/questions/8121575/po-files-and-gettext-vs-json-and-custom-i18n-library

@pcottle @aschrab @sid0 @remore @donkirkby

pcottle:
(closing out old issues, would still like to do this. I feel bad for the current translation process)

PR #516, opened 2018-12-03
Easy to multi-language, each language with one `.json` file

Hongarc:
This commit can help #161 .
If you want translate to another language, just  create `.json` file, and `require` it in `src/locales/index.js`

But it change very much (I can't count), so it may be wrong in some line, careful to merge it.
If you see any bug, please don't merge, just comment and I will fix it.


pcottle:
Hrm, what was your methodology for this? Did you do it in an automated fashion or by hand? :O

The root of #161 is that the translators for this app *also need to code* the translations, since they are stored in code. Most modern translation systems have a web UI instead where you can just type the translations you want and go from there -- the strings are stored in a local DB or resources file that the translator never needs to know about or edit.

I agree that you wont have to edit a central file to add a language in this format, but i actually think the workflow might be a bit worse. Some points:
 - First, this has introduced a number of short string IDs to map a given string from a level to the resources file. the issue with these strings is that they are brittle -- if you typo the string value, theres no immediate warning that you mistyped it and youd only catch it when testing the overall app. Contrast that to json where if you have duplicate keys youll get a JS hint error, or if the json is invalid you'll not be able to even run the app
 - Second, I think the workflow before was actually a bit easier. You can see the english string *right next* to where you type the translations; with this format you have to open up two different files and constantly scroll both of them as you translate the levels.
 - finally, its hard to see what languages exist for what levels yet. youd have to take the string ID and go search across N files to check N languages.

anyways I feel bad pushing back on all this work and I like the spirit of addressing #161 but I think this is a regression overall. next time youre planning big refactors just open an issue and we can discuss before :)

Hongarc:
I find where have multi-language.
And import it to one object (have `en_US`,`ko`,...) and export it to `.json` file, this step is auto.
And call require by hand (in `levels` and `dialog`), auto in `strings.js`

mexanga:
We can separate the values stored in the file `strings.js` by locales with this script :

```js
let values = {}
let locales = undefined
for (let key in strings) {
    if (undefined === locales) locales = Object.keys(strings[key])
    for (let locale of locales) {
        if (undefined === values[locale]) values[locale] = {}
        values[locale][key] = strings[key][locale];
    }
}
```

After the script executed, we get this output : 
image

We can retrieve the contents as JSON (it's already the case at the end of the script) and insert it into new files.

PR #518, opened 2018-12-05
Module

Hongarc:
I removed some module in `package.json`:
- `grunt-cli`: `grunt` include it and auto use match version for it, we can avoid some error if `grunt` call api `grunt-cli` don't support (example new `grunt`'s version has `foo` function but you forgot update `grunt-cli` )
- `events`: use `events` of Node.js, may be has some difference in Node.js 7 to 9, but `events` currently matches the Node.js 10.1 API
- `jquery`: your code never `require` `jquery`,  you use `jquery` from `jquery-1.8.0.min.js` (cnd or `lib/`)

*I see you have `lib/underscore` and you `require('underscore')` in source, You don't need add all of them, just use one*

Add `yarn.lock`.


pcottle:
great points!!

PR #519, opened 2018-12-05
Replace function of `underscore` with simple function

Hongarc:


pcottle:
Great! Did you test this out locally? just wondering since most of the animations and such dont have test coverage

Hongarc:
I was call:
````js
var debounce = function(func, time, immediate) {
    var timeout;
    return function() {
        var later = function() {
            timeout = null;
            if (!immediate) {
                func.apply(this, arguments);
            }
        };
        var callNow = immediate && !timeout;
        clearTimeout(timeout);
        timeout = setTimeout(later, time);
        if (callNow) {
            func.apply(this, arguments);
        }
    };
};

var throttle = function(func, time) {
    var wait = false;
    return function() {
        if (!wait) {
            func.apply(this, arguments);
            wait = true;

            setTimeout(function() {
                wait = false;
            }, time);
        }
    };
};
var mydebounce = debounce(()=>console.log('this is debounce'), 500);
var mythrottle = throttle(()=>console.log('this is throttle'), 500);
var i = 0;
var itv = setInterval(()=>{
    console.log(++i);
    mydebounce();
    mythrottle();
} , 99);

setTimeout(()=>{
    clearInterval(itv);
} , 1500);
````
Output:
````
1
this is throttle
2
3
4
5
6
7
this is throttle
8
9
10
11
12
13
this is throttle
14
15
this is debounce
````

I think it is ok.

PR #522, opened 2018-12-06
fix(css): fix offset top of select level (in firefox)

Hongarc:
In firefox, we can't see all of dialog select level (#474 )
old

After fixed it:
Preview result


pcottle:
awesome! thanks for fixing firefox, i only tested on chrome for mac which is bad...

Hongarc:
Thank you for your quick response

PR #523, opened 2018-12-07
Add log yarn, remove some line, what is covered by other line

Hongarc:


pcottle:
ah yeah i think I started with certain build files and then went to all. thx

Hongarc:
I think that you don't need some file in `.gitignore`, please clean them (you use some file in react-native,...etc, but you removed that feature)
Thanks.

PR #524, opened 2018-12-08
refactor(html): don't use underscore.min.js

Hongarc:
You use `underscore` with `var _ = require('underscore');` and haven't use `_` global, so you don't need use this script tag


pcottle:
cool! better TTI :D

PR #544, opened 2019-02-25
see #543 refined explanation of git describe

bwl21:
HI, I tried to improve description of git describe somehow.


src/levels/mixed/describe.js
```diff
@@ -1,4 +1,4 @@
-﻿exports.level = {
```
pcottle: whats this change for?

pcottle:
gonna merge but will check this works :)

pcottle:
seems to work!

src/levels/mixed/describe.js
bwl21: It marks this line as deleted and added. I could not find out what has been changed in this line. Maybe it is something with whitespace ...

PR #546, opened 2019-02-26
Make contributions easier

jankeromnes:
👋 Hello!

Thanks for making this cool Git tutorial. I work on gitpod.io, a service that provides free online workspaces for open source projects, and I thought you might be interested in having pre-configured dev environments to help your contributors get started easily.

I've configured Gitpod to pre-install `grunt` and all `npm` dependencies, then to automatically build your app and open `index.html` in a preview pane. Here is what it looks like:

[Open in Gitpod](https://gitpod.io/#https://github.com/jankeromnes/learnGitBranching/blob/master/src/js/git/index.js)
screenshot 2019-02-26 at 17 45 20
Please have a look when you have time. 🙂

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,0 +1,6 @@
+ports:
+- port: 8000
+  onOpen: open-preview
+tasks:
+- init: npm install && npm install -g grunt-cli
+  command: grunt build && python3 -m http.server 8000
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ You'll also need `npm` to download all the dependencies of the project.
 
 The general workflow / steps are below:
 
-```
+```bash
 git clone <your fork of the repo>
 cd learnGitBranching
 npm install # to install all the node modules I depend on
@@ -96,6 +96,10 @@ git commit -am "My new sweet feature!"
 git push
 # go online and request a pull
 ```
+
+Alternatively, you can also build and run the app in a pre-configured online workspace:
+
+[![Open in Gitpod](https://gitpod.io/button/open-in-gitpod.svg)](https://gitpod.io/#https://github.com/pcottle/learnGitBranching/blob/master/src/js/git/index.js)
 
 ## Helpful Folks
 A big shoutout to these brave souls for extensively testing our sandbox and finding bugs and/or inconsistencies:


README.md
```diff
@@ -77,7 +77,7 @@ You'll also need `npm` to download all the dependencies of the project.
 
 The general workflow / steps are below:
 
-```
+```bash
```
jankeromnes: Sorry, this is a drive-by change: I thought the setup instructions could be more readable when syntax-highlighted as Bash.

pcottle: nah this is a good change!

.gitpod.yml
```diff
@@ -0,0 +1,6 @@
+ports:
+- port: 8000
+  onOpen: open-preview
+tasks:
+- init: npm install && npm install -g grunt-cli
+  command: grunt build && python3 -m http.server 8000
```
pcottle: if you're ok with doing hard refreshes, `grunt fastBuild` is a faster version. it skips the hashing step and simply puts the assets in build.js or something, which means you need to hard refresh but saves you a decent chunk of time

README.md
```diff
@@ -97,6 +97,10 @@ git push
 # go online and request a pull
 ```
 
+Alternatively, you can also build and run the app in a pre-configured online workspace:
+
+[![Open in Gitpod](https://gitpod.io/button/open-in-gitpod.svg)](https://gitpod.io/#https://github.com/pcottle/learnGitBranching/blob/master/src/js/git/index.js)
```
pcottle: oh so the link is the file that opens, but it reads the yaml file automatically?

pcottle:
I'm super down to accept this, I just think we need some kind of callout on how to *rebuild* the project. is that included as part of `command`? also see the inline about the `fastBuild` variant

pcottle:
actually this provides value already, but curious to hear what you think about my comments above

jankeromnes:
Hi @pcottle, thanks for the fast review! I'm so happy that you find it useful. 😄 You also raised some great points, I'll reply to your other comments individually.

.gitpod.yml
jankeromnes: Good idea! I'll follow up to use `grunt fastBuild` instead of `grunt build`. Saving time is 👌.

README.md
jankeromnes: Yes, after the `gitpod.io/#` prefix, you can use basically any GitHub URL (specific file, branch, pull request...) and it will try to do the right thing. It will also look for a `.gitpod.yml` config file at the root of the repository, to see if there is anything specific to do to set up that project.

I chose to pre-open `src/js/git/index.js` in the editor, but feel free to change it if you have a better idea.

jankeromnes:
> I'm super down to accept this, I just think we need some kind of callout on how to _rebuild_ the project. is that included as part of `command`? also see the inline about the `fastBuild` variant

That's a great question. Maybe we can improve what Gitpod does beyond the immediate project setup.

Basically, when you start a workspace for the first time, it will run `init` and then `command`; and if you restart a workspace that already existed, it will just run `command`.

In Gitpod, I usually use `Ctrl`+`C` and `UpArrow` in the Terminal to interrupt and re-run the `command`, but maybe that's not obvious, or there can be better ways to suggest relevant commands.

A similar idea was to allow configuring various helpful project-specific workflows in `.gitpod.yml`, and to show these in the IDE, e.g. via a `Run` button or dropdown. It's filed as https://github.com/gitpod-io/gitpod/issues/247 but hasn't been implemented yet. Maybe I should take another look there.

jankeromnes:
Actually, maybe there is a simpler way to call out how to rebuild: What about adding something like `echo "To rebuild the app, run 'grunt fastBuild' and reload 'index.html'."` to the `command`?

pcottle:
Oh yeah that'd be a nice hack/fix to give people the heads up!

PR #547, opened 2019-02-26
Update German translation

MichaWiedenmann:
Update translation of "Show Goal" and "Hide Goal".

### DIFF
--- a/src/js/intl/strings.js
+++ b/src/js/intl/strings.js
@@ -1169,7 +1169,7 @@ exports.strings = {
   'show-goal-button': {
     '__desc__': 'button label to show goal',
     'en_US': 'Show Goal',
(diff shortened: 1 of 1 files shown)


pcottle:
```
gpg: Signature made Tue 18 Dec 2018 04:04:55 PM UTC using RSA key ID B6FF4DE3
gpg: BAD signature from "Yarn Packaging <yarn@dan.cx>"
> GPG signature for this Yarn release is invalid! This is BAD and may mean the release has been tampered with. It is strongly recommended that you report this to the Yarn developers.
The command "curl -o- -L https://yarnpkg.com/install.sh | bash" failed and exited with 1 during .

```

supposedly this caused the travis build to fail but im 99% certain its just a flaky error. ill repush though

PR #548, opened 2019-02-27
Use 'grunt fastBuild' in Gitpod to build the app faster

jankeromnes:
Follow-up improvement to #546.

@pcottle I'm also happy to add an `echo` command as suggested in https://github.com/pcottle/learnGitBranching/pull/546#issuecomment-467761064 if you think that's a good idea.

Please take a look when you have time.


pcottle:
yeah lets do the echo. youll need to do it before the python server startup though since that command never finishes, or youll need to send it to the background with `&` or something.

PR #549, opened 2019-02-28
Print helpful welcome message in Gitpod

jankeromnes:
This implements https://github.com/pcottle/learnGitBranching/pull/546#issuecomment-467761064 as requested in https://github.com/pcottle/learnGitBranching/pull/548#issuecomment-467972574.

Here is what it looks like (see the two additional lines printed in the Terminal):

[Open in Gitpod](https://gitpod.io/#https://github.com/jankeromnes/learnGitBranching/blob/master/src/js/git/index.js)
screenshot 2019-02-28 at 12 51 14

Please let me know if you like the wording, or how it can be improved.

Notes:
- I used `printf` instead of `echo` as it's generally safer with escaped characters and CLI options across various shells
- I've disabled `python3 -m http.server 8000`'s error output, as it was printing new lines on every served resource, thus quickly hiding the message

Another option could be to open two terminals side-by-side, and have e.g. the left terminal print the message, and the right terminal running the server. For example, the Firefox Devtools Profiler also did [something like that](https://github.com/firefox-devtools/profiler/blob/master/.gitpod.yml#L7-L8).

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -3,4 +3,7 @@ ports:
   onOpen: open-preview
 tasks:
 - init: npm install && npm install -g grunt-cli
-  command: grunt fastBuild && python3 -m http.server 8000
+  command: >
+    grunt fastBuild &&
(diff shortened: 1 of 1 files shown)


pcottle:
This looks great!

PR #552, opened 2019-03-11
Add configuration for Galician language

viticlick:
Add translation for [galician language](https://en.wikipedia.org/wiki/Galician_language)


pcottle:
Awesome thanks so much for this huge amount of work! I’ll try to push the updates this week

PR #558, opened 2019-04-10
fix(readme): hide comment in README.md

Hongarc:
image

Hide `[//]: contributor-faces` in `README.md`


pcottle:
lol i couldnt even do this correctly... thanks!

pcottle:
hrm, seems theres a conflict now :-/

pcottle:
there we go

Hongarc:
And I think you add image contributor under Gitpod is not a good idea, I think we should add it under **Helpful Folks**

PR #563, opened 2019-04-18
Explicitly fail when unsupported range operator is specified

eatdrinksleepcode:
Currently the range operator (e.g. "master..future") is misinterpreted as a relative ref, due to a regex which strips off everything from the dots onwards. As a result, "master..future" is interpreted as just "master". Because the regex is not strict, many other illegal forms are also misinterpreted as relative refs, such as "master&&&".

This PR adds "$" (zero-length line terminator) to the end of the regex, forcing the regex to match the whole argument. This way, if the argument is not actually a relative ref, it will not be interpreted as such. The result will be an error ("The ref master..future does not exist or is unknown"), which I think is preferable to the current behavior.

See #561

### DIFF
--- a/__tests__/git.spec.js
+++ b/__tests__/git.spec.js
@@ -142,6 +142,13 @@ describe('Git', function() {
     );
   });
 
+  it('Range operator is not supported', function() {
+    expectTreeAsync(
+      'git checkout -b side C0; git cherry-pick C1..C0',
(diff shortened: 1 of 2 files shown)


pcottle:
Thanks for adding test coverage as well! I definitely should have had this in there since day one. I might have been trying to handle trailing whitespace or something clowny, but that's better handled in the arguments parsing

PR #564, opened 2019-04-19
chore: use `every` instead of `uniq`

Hongarc:
You use `_.uniq` but don't return value of `_.uniq`.

You use that function to check `result` value so my recommendation is to use `every`, it will break if the value of callback `false` to avoid call other callback when `result` is `false`

And I remove one line call `underscore` but don't call `_`


pcottle:
Oh yeah this is much cleaner! Also matches the `C\every` from hack (which I use at work)

PR #565, opened 2019-04-19
More robust support for revision ranges

eatdrinksleepcode:
This is the first step in adding greater support for revision ranges (see #561). Currently, only log supports revision ranges, it only supports specific kinds (single tip includes and excludes), and only in very limited formats ("git log include" or "git log include ^exclude"). With this PR, log now supports any number of single tip includes or excludes, in any order, as real git does.

This PR also introduces support for the "rev-list" command, which uses the same revision range capability. In addition to increased git compatibility, the rev-list command also provides much simpler output with which to test the output ordering of revision ranges.

Compatibility Note: this PR changes the output ordering of the log command. The previous ordering was essentially a breadth-first search. The new ordering follows the git default, which is based on commit date.

### DIFF
--- a/__tests__/base.js
+++ b/__tests__/base.js
@@ -1,3 +1,5 @@
+var Q = require('q');
+
 var HeadlessGit = require('../src/js/git/headless').HeadlessGit;
 var TreeCompare = require('../src/js/graph/treeCompare.js');
 
@@ -95,6 +97,27 @@ var expectLevelSolved = function(levelBlob) {
   expectLevelAsync(headless, levelBlob);
 };
 
+var runCommand = function(command, resultHandler) {
+  var headless = new HeadlessGit();
+  var deferred = Q.defer();
+  var msg = null;
+
+  deferred.promise.then(function(commands) {
+    msg = commands[commands.length - 1].get('error').get('msg');
+  });
+
+  runs(function() {
+    headless.sendCommand(command, deferred);
+  });
+  waitsFor(function() {
+    if(null == msg) {
+      return false;
+    }
+    resultHandler(msg);
+    return true;
+  }, 'commands should be finished', 500);
+};
+
 var TIME = 150;
 // useful for throwing garbage and then expecting one commit
 var ONE_COMMIT_TREE = '{"branches":{"master":{"target":"C2","id":"master"}},"commits":{"C0":{"parents":[],"id":"C0","rootCommit":true},"C1":{"parents":["C0"],"id":"C1"},"C2":{"parents":["C1"],"id":"C2"}},"HEAD":{"target":"master","id":"HEAD"}}';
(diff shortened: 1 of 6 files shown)


__tests__/base.js
```diff
@@ -95,6 +97,27 @@ var expectLevelSolved = function(levelBlob) {
   expectLevelAsync(headless, levelBlob);
 };
 
+var runCommand = function(command, resultHandler) {
+  var headless = new HeadlessGit();
+  var deferred = Q.defer();
+  var msg = null;
+
+  deferred.promise.then(function(commands) {
+    msg = commands[commands.length - 1].get('error').get('msg');
```
pcottle: i vaguely remember why i store all command output in the error field, but its certainly silly

__tests__/git.spec.js
```diff
@@ -275,4 +276,104 @@ describe('Git', function() {
...
+      it('multiple excluded revisions', function() {
+        runCommand(SETUP + 'git rev-list all ^right ^left', function(commandMsg) {
+          expect(commandMsg).toBe('C6\n');
+        });
+      });
+    });
+  });
+
+  describe('Log supports', function() {
+    var SETUP = 'git co -b left C0; gc; git merge master; git co -b right C0; gc; git merge master; git co -b all left; git merge right; ';
```
pcottle: sometimes i use serialized tree states for this type fo setup, but this works as well!

src/js/git/headless.js
```diff
@@ -117,14 +119,15 @@ HeadlessGit.prototype.sendCommand = function(value, entireCommandPromise) {
...
+      commands.push(commandObj);
       return thisDeferred.promise;
     }.bind(this));
   }, this);
 
   chain.then(function() {
     var nowTime = new Date().getTime();
     if (entireCommandPromise) {
-      entireCommandPromise.resolve();
+      entireCommandPromise.resolve(commands);
```
pcottle: oh so this adds a return arg to the promise? good call

src/js/git/index.js
```diff
@@ -3079,6 +3068,79 @@ var Tag = Ref.extend({
...
+  specifiers.forEach(function(specifier) {
+    if(self.isExclusion(specifier)) {
+      exclusions.push(specifier.slice(1));
+    } else {
+      inclusions.push(specifier);
+    }
+  });
+
+  exclusions.forEach(function(exclusion) {
+    self.addExcluded(Graph.getUpstreamSet(self.engine, exclusion));
```
pcottle: glad you found these helpers

src/js/models/commandModel.js
```diff
@@ -151,6 +147,12 @@ var Command = Backbone.Model.extend({
     this.validateArgBounds(args, 0, 2);
   },
 
+  impliedHead: function(args, min) {
```
pcottle: 👍

pcottle:
Pushed the site with these changes, it should be live :)

__tests__/git.spec.js
eatdrinksleepcode: Yeah, serialized tree state would probably be more in keeping with the way the other tests are setup. I'll be perfectly honest: just entering the commands was easier :)

PR #567, opened 2019-04-21
Trim

Hongarc:
- Use `Object.assign` instead of module `assign`
- Remove all white space at end of line


pcottle:
Looks great! I think I had this before the polyfill days but prefer this for sure

PR #569, opened 2019-04-22
Range operator

eatdrinksleepcode:
This PR adds support for the range operator ("branchA..branchB") to the `log` and `rev-list` commands.

It also refactors revision specifier processing into a "pipeline" model, to make it easier to support new kinds of specifiers.

See #561

### DIFF
--- a/src/js/git/index.js
+++ b/src/js/git/index.js
@@ -3070,39 +3070,49 @@ var Tag = Ref.extend({
 
 function RevisionRange(engine, specifiers) {
   this.engine = engine;
-  this.included = {};
-  this.excluded = {};
+  this.tipsToInclude = [];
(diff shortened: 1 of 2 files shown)


src/js/git/index.js
```diff
@@ -3070,39 +3070,62 @@ var Tag = Ref.extend({
 
 function RevisionRange(engine, specifiers) {
```
pcottle: this doesnt have a ton of dependencies so maybe we could split it out one day, but no big deal.

pcottle:
Awesome work @eatdrinksleepcode! 🎉

src/js/git/index.js
eatdrinksleepcode: yeah makes sense

PR #572, opened 2019-04-23
Update test in travis-ci

Hongarc:
Check it will build or not.


pcottle:
Thanks! this is better :)

PR #575, opened 2019-05-01
Update yarn.lock file

eatdrinksleepcode:
A number of recent commits have updated versions in package.json without
updating the yarn.lock file. This commit brings the yarn.lock file up to date.


pcottle:
go for it, i kept resetting this thinking it was a change I shouldnt commit >___<

Issue #574, opened 2019-04-29
yarn vs npm


eatdrinksleepcode:
It appears that the preferred package manager was changed from npm to yarn late in 2018 (see #513 and #517). However, references to npm still exist in a couple of places:

1. The README, which still shows how to use npm to install dependencies: https://github.com/pcottle/learnGitBranching/blob/ac880dad0f3b7b2cc499150a828f0842c6a32520/README.md#building-yourself--contributing-functionality
This should presumably be changed to refer to yarn instead of npm.

2. The .gitpod.yml file which was added in February: https://github.com/pcottle/learnGitBranching/blob/ac880dad0f3b7b2cc499150a828f0842c6a32520/.gitpod.yml#L5
This should presumably be changed to use yarn instead of npm.

3. The grunt file, which uses the `loadNpmTasks` grunt method, and has the comment "all my little npm helpers": https://github.com/pcottle/learnGitBranching/blob/ac880dad0f3b7b2cc499150a828f0842c6a32520/Gruntfile.js#L228-L237
The method is IMHO unfortunately named, but nothing can be done about that. Not sure what if anything should be done with the comment.

4. The `.gitignore` file includes `npm-debug.log`: https://github.com/pcottle/learnGitBranching/blob/ac880dad0f3b7b2cc499150a828f0842c6a32520/.gitignore#L2
No strong feeling about this, probably ok to leave it.

pcottle:
Good point! The README is probably the most important update but the others are good for consistency

@jankeromnes does gitpod work with yarn as well? I'll assume so

`loadNpmTasks` really is an unfortunate name. Maybe someone can submit a PR to add an alias like `loadNodeModuleTasks`...

ill update these other ones though.

pcottle:
heads up to @Hongarc

Hongarc:
@eatdrinksleepcode  Feel free to create a PR(if not, I will try to resolve this issue to night +7GMT)

eatdrinksleepcode:
Was intending to create a PR but I see you beat me to it :) Just added one more

jankeromnes:
> @jankeromnes does gitpod work with yarn as well? I'll assume so

It does! In fact we even use `yarn` for Gitpod itself. 🙂 (And apologies for the late reply, I was on vacation last week.)

[...]

PR #576, opened 2019-05-01
Update README reference from npm to yarn

eatdrinksleepcode:
See #574


pcottle:
i knew id miss something :P

PR #577, opened 2019-05-01
Check that there are no source file changes before building

eatdrinksleepcode:
If dependencies are changed (in package.json) without updating the yarn.lock
file, or if any other part of the build process starts producing unignored
files, the CI build will fail, so that the problem can be corrected before the
PR is merged.

See #575

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,4 +11,6 @@ cache:
   directories:
     - "node_modules"
 script:
+  - ./checkgit.sh "Source files were modified before build; is yarn.lock out of sync with package.json?" || travis_terminate $?
   - yarn grunt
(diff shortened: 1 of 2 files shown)


pcottle:
Hrm maybe to debug the build failure we can output the status of `git status` before this line? also to aid in debuggability

eatdrinksleepcode:
I updated with a script that will run git status and output a message.

See https://travis-ci.org/pcottle/learnGitBranching/jobs/526793367 and https://travis-ci.org/pcottle/learnGitBranching/jobs/526793368 for example output.

pcottle:
There we go! Awesome :D

pcottle:
Thanks so much for making our CI pipeline better! really cool to catch these things with travis

PR #579, opened 2019-05-02
Use `babelify` instead of `grunt-react`

Hongarc:
- [DEPRECATED] `grunt-react` and they notice `Please use grunt-babel instead`
- But you use `grunt-react` for transform `browserify` so I use `babelify` instead


pcottle:
thanks for updating the yarn lock too :)

PR #580, opened 2019-05-07
Use `yarn grunt` instead of the global `grunt`

jankeromnes:
This change fixes the the Gitpod demo by setting the `yarn` global prefix to `/workspace/.yarn`, and adding it to the `PATH` env variable.

Follow-up to https://github.com/pcottle/learnGitBranching/issues/574#issuecomment-490173208.

Note: The `before` step runs both during Gitpod's first workspace start (`before && init && command`) and subsequent restarts (`before && command`). More info [here](https://www.gitpod.io/docs/44_Config_Start_Tasks/#code-classlanguage-textbeforecode-command).

[Open in Gitpod](https://gitpod.io/#https://github.com/jankeromnes/learnGitBranching/blob/patch-1/src/js/git/index.js)

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -2,8 +2,8 @@ ports:
 - port: 8000
   onOpen: open-preview
 tasks:
-- init: yarn install && yarn global add grunt-cli
+- init: yarn install
   command: >
-    grunt fastBuild &&
-    printf "\nWelcome to Learn Git Branching\nTo rebuild the app, simply run 'grunt fastBuild' and reload index.html.\n\n" &&
+    yarn grunt fastBuild &&
+    printf "\nWelcome to Learn Git Branching\nTo rebuild the app, simply run 'yarn grunt fastBuild' and reload index.html.\n\n" &&
     python3 -m http.server 8000 2>/dev/null
--- a/README.md
+++ b/README.md
@@ -85,13 +85,13 @@ yarn install
 
 git checkout -b newAwesomeFeature
 vim ./src/js/git/index.js # some changes
-grunt fastBuild # skips tests and linting, faster build
+yarn grunt fastBuild # skips tests and linting, faster build
 
 # after building you can open up your browser to the index.html
 # file generated and see your changes
 
 vim ./src/js/git/index.js # more changes
-grunt build # runs tests and lint
+yarn grunt build # runs tests and lint
 
 git commit -am "My new sweet feature!"
 git push


.gitpod.yml
```diff
@@ -2,7 +2,8 @@ ports:
 - port: 8000
   onOpen: open-preview
 tasks:
-- init: yarn install && yarn global add grunt-cli
+- before: yarn config set prefix /workspace/.yarn && export PATH="$PATH:`yarn global bin`"
+  init: yarn install && yarn global add grunt-cli
   command: >
     grunt fastBuild &&
```
pcottle: yeah wanna do `yarn grunt fastBuild` down here?

.gitpod.yml
```diff
@@ -2,7 +2,8 @@ ports:
 - port: 8000
   onOpen: open-preview
 tasks:
-- init: yarn install && yarn global add grunt-cli
+- before: yarn config set prefix /workspace/.yarn && export PATH="$PATH:`yarn global bin`"
```
pcottle: and we can probably rm all of this?

.gitpod.yml
```diff
@@ -2,7 +2,8 @@ ports:
...
+  init: yarn install && yarn global add grunt-cli
```
pcottle: and this becomes just `yarn install`

pcottle:
in the future, is there an easy way to test these alternate gitpod files?

jankeromnes:
@Hongarc @pcottle I've implemented your suggestions. Looks much cleaner now, thanks!

> in the future, is there an easy way to test these alternate gitpod files?

Yes, there are several ways to test `.gitpod.yml` changes:

- Fork the repository, change `.gitpod.yml`, open fork in Gitpod (e.g. by prefixing the repo URL with `gitpod.io/#`)
- Make a pull request that changes `.gitpod.yml`, open the pull request in Gitpod (same prefix)
- Install the [Gitpod app](https://github.com/apps/gitpod-io) for this repo, which adds a Check to pull requests, and continuously builds workspaces on changes

The last one is probably the easiest. 🙂

Hongarc:
```
printf "\nWelcome to Learn Git Branching\nTo rebuild the app, simply run 'grunt fastBuild'" 
```
You should add `yarn` before `grunt` to avoid someone create an issue that error when run `grunt fastBuild`

jankeromnes:
> You should add `yarn` before `grunt` to avoid someone create an issue that error when run `grunt fastBuild`

Ah, right, sorry for the oversight. 😓 Thanks for catching it!

jankeromnes:
> And can you change them in `README.md`? (`grunt fastBuild` in `README.md`)
> 
> I think we need to resolve it in this PR.

Sure! I also saw that when grepping for remaining `grunt` calls.

However, I wasn't sure what to do with the paragraph above that tells you to install `grunt` globally:

> That means you'll need the "Grunt.js" build tool to build the app:
> 
> https://gruntjs.com/getting-started

Hongarc:
We don't need to install global, because we install it in `devDependencies`, and don't need `grunt-cli`, because `grunt"0.4.5` don't have `bin` file, but `grunt"1.0.4` have it.

pcottle:
Looks great now! thanks so much :D

PR #582, opened 2019-05-16
Set line height of icons bar icons

### LINKED ISSUE #581: Incorrect alignment of icons bars in Firefox
In Firefox, the icons bars that show when a command has been executed are aligned incorrectly. See below image.

This is due to the Font Awesome icons used in this bar getting an incorrect computed `line-height` of `19.2px`, overflowing the `18px` high `span` containers, causing them to take up more vertical space than the containing `div` and preventing them from floating all the way to the right.

It can be fixed by manually setting the `line-height` of the icons `18px` or below.

Learn Git Branching Firefox Visual Bug

kimsey0:
Fixes #581.

### DIFF
--- a/src/style/main.css
+++ b/src/style/main.css
@@ -497,6 +497,7 @@ div.reactCommandView p.commandLine span.icons i:first-child {
 
 p.commandLine span.icons i {
   opacity: 0;
+  line-height: 1em;
 }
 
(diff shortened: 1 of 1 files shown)


pcottle:
Awesome! Thanks so much for submitting the fix

pcottle:
Should be pushed out to the site now :)

PR #584, opened 2019-06-05
feat: make es_ES available from the languages menu

### LINKED ISSUE #528: Spanish translation is too localized
The translation for the Spanish locale is too localized. Instead of using international, neutral Spanish it's riddled with argentinean expressions, particularly regarding imperatives (_"Aprendé"_ instead of _"Aprenda"_ or _"Tenés"_ instead of _"Tienes"_).

Splaktar:
rename es_AR to argentino
 but leave it as the default for Spanish
 since es_ES is not fully translated

Fixes #528


pcottle:
good callout! thx for the PR

PR #599, opened 2019-09-04
added git tag -d <tag> and  fixed error message

LukasSchmid97:
Added the tag-deleting feature and fixed a bug, where the duplicate tag-name wouldn't show in the error message


pcottle:
wow looks awesome!! thanks a ton :D

PR #602, opened 2019-09-16
Change upper value in 'git-error-args-few' string to lower

mexanga:
The name of the branch doesn't correspond at all with the commit. The "french git rebase" has been fixed in replacing a value in the strings by the good one.
When typing "git rebase" in the command line, we have to refresh the page to reuse the application.


pcottle:
Good catch 🎉

PR #603, opened 2019-09-17
Fix the level mixed1 : bugFix on C4' was optionnal.

mexanga:
Now we have to get the branches 'master' and 'bugFix' on the commit C4' to passing the level.

### DIFF
--- a/src/levels/mixed/grabbingOneCommit.js
+++ b/src/levels/mixed/grabbingOneCommit.js
@@ -5,6 +5,11 @@ exports.level = {
       function(data) {
         return data.C4 > data.C1;
       }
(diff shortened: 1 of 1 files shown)


pcottle:
Can you clarify what solution was correct (but not seen as such) without this fix? or what state you want to prevent asserting from?

mexanga:
We can do the exercise with "git checkout master; git cherry-pick bugFix" but the branch 'bugFix' isn't on C4' and thus, doesn't correspond with the tree given.

pcottle:
Oh my point is that I want your command to actually work -- I dont think we should prevent that solution!

PR #605, opened 2019-09-19
Update index.js

Zabari:
this is a HUGE missed opportunity for a pun, and I think the pun fits better with the tone of the site generally


pcottle:
Haha true! To make this not look like a typo, could we do:

```
      'en_US': '"Git" comfortable with modifying the source tree :P',
```

Zabari:
Great idea! I updated it :D

pcottle:
Perfect! 🎉

PR #608, opened 2019-09-24
fix: force `ls` and `cd` end with space

Hongarc:
I type `lssss` in command line, it will think that is `ls`. it should be say `isn't supported`

### DIFF
--- a/src/js/sandbox/commands.js
+++ b/src/js/sandbox/commands.js
@@ -15,12 +15,12 @@ var Warning = Errors.Warning;
 var CommandResult = Errors.CommandResult;
 
 var instantCommands = [
-  [/^ls/, function() {
+  [/^ls /, function() {
     throw new CommandResult({
       msg: intl.str('ls-command')
     });
   }],
-  [/^cd/, function() {
+  [/^cd /, function() {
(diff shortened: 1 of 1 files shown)


src/js/sandbox/commands.js
```diff
@@ -15,12 +15,12 @@ var Warning = Errors.Warning;
 var CommandResult = Errors.CommandResult;
 
 var instantCommands = [
-  [/^ls/, function() {
+  [/^ls /, function() {
```
waldyrious: It would be better to use `/^ls\b` instead, I think. `\b` in regex represents a word boundary, which is more explicit in meaning than a space.

pcottle:
With this change, wouldn't we need `ls` need to have a trailing space in order for the command to run? I tried it out and I got that result:
Screen Shot 2019-09-25 at 10 38 39 AM

as @waldyrious  mentioned, this would be better with `\b` so we just enforce theres a boundary. but that also means you can add content after `ls` but thats ok. lets go with that change

pcottle:
Screen Shot 2019-09-25 at 10 40 30 AM

PR #610, opened 2019-09-25
fix: get `ls` and `cd` with right way

Hongarc:
#608 

Before:
image

- Why don't use `\b`: `ls-test`

I checked `$` in end of line

After: 

image


Hongarc:
@pcottle Please give me feedback!

Thank you.

pcottle:
Oh sorry I missed this! Yeah this is better since we don't want all boundary characters here, only spaces or the end of the line.

typically in english you could be looking at stuff like "mid word," or "final word." where the `.` and `,` are fine boundaries (if you're searching for "word") but thats not the case here. thanks for the PR

PR #611, opened 2019-10-03
Disable "git fakeTeamwork" in push level (remote6)

PhilipEnchin:
As it stands, level remote6 (Git Pushin') can be solved without using `git push` at all:

```
$ git fakeTeamwork 2
$ git pull
```


pcottle:
Oh man good point! 🙀

PR #612, opened 2019-10-03
Disable "git branch" and "git checkout" in source nothing level (remoteAdvanced7)

PhilipEnchin:
Despite the hint to the contrary, `git branch` is not disabled here. Since `git checkout -b` also works here, I disabled them both.


pcottle:
Sure! I usually only disable these commands when peoples first reaction to solving the level is to using another set of commands (rather than what I explain in the lesson) but its fair to disable these and keep people on track

PR #616, opened 2019-10-22
Change where HEAD points to match objective

dzampino:
The instructions for level rampup1 say to "...detach HEAD from bugFix and attach it to the commit instead..." HEAD in this level points to the master branch at the start. I changed the starting target of HEAD from master to bugFix to reflect these instructions.


pcottle:
Yeah good point. I think my original intent here was to have more of a visible animation as `HEAD` moves from C2 to C4 (rather than the more subtle change) but I think people will learn it either way

dzampino:
I totally agree with your reasoning; I just went with an easier fix that doesn't require new translations.

PR #622, opened 2019-12-18
Es es translation

flyingbot91:
This PR includes the **es_ES** (Spain) translations.

Just a quick question:

I would like to add the **es_ES** language in the file _LocaleStore.js_, but I think this might cause conflicts with the Argentinian **es_AR** key. How can I fix this?

See below the variable I am talking about:

```
var langLocaleMap = {
  en: 'en_US',
  zh: 'zh_CN',
  ja: 'ja',
  ko: 'ko',
  es: 'es_AR',
  fr: 'fr_FR',
  de: 'de_DE',
  pt: 'pt_BR',
  ru: 'ru_RU',
  uk: 'uk'
};
```
Thank you for your help.


pcottle:
Oh for that map, it just means if we receive the language from the browser as `es` we will automatically map it into `es_AR`. I presume `es_ES` is more of a generalized / more popular locale for `es` language users? if so, feel free to replace it in that mapping.

its just used here:
https://github.com/pcottle/learnGitBranching/blob/e92fe2c45e4de174126940b176352b562f1e3a09/src/js/stores/LocaleStore.js#L48

to jump into that specific locale from the language

pcottle:
I'll merge this for now but feel free to add another PR for that entry in the map!

pcottle:
works!
https://learngitbranching.js.org/?locale=es_ES

😇

PR #624, opened 2020-01-02
Added "es-ES" locale

flyingbot91:
The modification in the PR will be used when the browser locale (ordered variable) has been set in a way like this (or similar):

`>>> console.log(navigator.languages);`
`>>> Array(4) [ "es-ES", "es", "en-US", "en" ]`

PR #625, opened 2020-01-02
Other version of PR #624 -- map header directly from browser

pcottle:
lets see


flyingbot91:
Now it works.
I skipped the tests before. My bad.

PR #635, opened 2020-02-05
Add RUS translate for "Locked Master" level

andreysy4:
Added Russian translation with slight design changes.
PS It was a good practical lesson :)))


pcottle:
Awesome! Every bit helps :)

PR #652, opened 2020-02-29
Korean translation of lockedMaster.js

Shubidumdu:
While I was learning to use git with this app, I found a level(lockedMaster) that has not been translated yet, so I translated it in Korean. :) this app helped me a lot!


pcottle:
This is awesome! It's a pretty new level so that's why it's not translated yet -- thanks @Shubidumdu for the contribution :D

pcottle:
Should be live on the site then!

PR #653, opened 2020-03-05
No grep

Hongarc:
Origin:
 - Use `grep` to detect string `intl.str`
 - Cannot make sure find exactly by current regex
 - ...

To fix it:
 - I use the `fs` module to read the text of the file
 - `join` of `path` module to choose the true path
 - Find text with new regex: `intl.str` to `intl\.str`, add `\s*` between name of function and text
 - Did not call key many times with Set
 - Show unused key
image

*Note: remove some function unused*


pcottle:
Awesome! Thanks so much for fixing this up -- I haven't run it in years but I'd rather have it work :P

Issue #654, opened 2020-03-05
Upgrade module


Hongarc:
I attempt to upgrade the module of this module.

However, too many `grunt-plugin` is outdated(install without `yarn.lock` have error)

In my opinion, `gulp` is easy to understand, do you want to use it? (I made a config file for `gulp` from your `Gruntfile.js`, have an error with old version `jasmine`)

You create a config `jshint` in `Gruntfile.js`, how about create `.jshintrc` file(it can work with a plugin of your IDE)

I use eslint and other plugins(have many options to lint code)

What is your choice?
- [ ] Grunt
- [ ] Gulp
- [ ] Eslint
- [ ] Jshint
- [ ] Update jasmise
- [ ] Keep jasmise

I create this issue because I install and test fails with `npm i` ...

Thank you.

pcottle:
I'd be totally willing to switch to Gulp -- thats more popular and more modern now right?

Don't know how much work it is to switch over. But yeah I think at this point we only support `yarn` and not `npm`. Ill update the readme

pcottle:
(Oh we already reference yarn in the readme, yay)

Hongarc:
Can you help me update `jasmine`?

I need a little time to understand that test and migrate to the new version.

I am going to create a pull request without using `jasmine`(until I migrate it to newversion)

PR #655, opened 2020-03-06
[feat] update url if user change locale

Hongarc:
Add locale in `href`

If the user turns off the browser and opens it again, they keep their's locale

Oh, build have error(I cannot test in my local because I am attempting #654)

Wait for me fix this Travis error

### DIFF
--- a/src/js/stores/LocaleStore.js
+++ b/src/js/stores/LocaleStore.js
@@ -2,6 +2,7 @@
 
 var AppConstants = require('../constants/AppConstants');
 var AppDispatcher = require('../dispatcher/AppDispatcher');
+var util = require('../util');
 var EventEmitter = require('events').EventEmitter;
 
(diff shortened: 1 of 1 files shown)


pcottle:
Oh wait there’s one test failure. Can we make the lookup of document null safe so it’ll pass in a node environment?

‘’’
Failures:
623
624  1) LocaleStore changes locales
625   Message:
626     ReferenceError: document is not defined
627   Stacktrace:
628     ReferenceError: document is not defined
629  at Object.changeLocale

Hongarc:
> Will this trigger if we are just changing the locale from a browser header though?

@pcottle Yes, I updated it to auto-detect all case

pcottle:
Even better! thanks doing the util browser check -- wanted to keep those tests healthy

pcottle:
(I'll push the site with these changes shortly when I'm on my home computer)

PR #658, opened 2020-03-15
fix(typo): remove unnecessary `.`

Hongarc:
It is better with `;` or `.`, not both.

Or you can use `.` instead of `;` ( same the normal sentence)


pcottle:
perfect! thanks

PR #664, opened 2020-03-27
add french translation to lockedMaster

cbobinec:
Hi,
Here is a french translation of the level "Remote 8 locked master". Other translations are missing, sorry I'm not qualified ;-)


pcottle:
Thanks again!

PR #661, opened 2020-03-24
chore(test): mock `localStorage`

Hongarc:
I know that you are very busy, just resolve your problem.

Recently, I create the chore pull request(not the important change) so you can temporarily ignore it in this time.

PR #669, opened 2020-03-30
fix: keep `localStorage`

Hongarc:
Related #661

I assign `var localStorage` inside if-block(not called) so `localStorage` doesn't refer to `window.localStorage`.

image

BTW, fix the report by JSHint.


pcottle:
🎉

PR #670, opened 2020-03-30
fix: `CommandView` don't need to unmount itself

Hongarc:
When you destroy command with [`toDestroy[i].destroy();`](https://github.com/pcottle/learnGitBranching/blob/699d5ffb71f197885a7a585b88ce11546d04d6c8/src/js/react_views/CommandHistoryView.jsx#L94), `CommandHistoryView`'s props will 
 change and render its child(`CommandView`)

In my opinion, we don't need to unmount it(react do it)
and pass this error
image

### DIFF
--- a/src/js/react_views/CommandView.jsx
+++ b/src/js/react_views/CommandView.jsx
@@ -15,17 +15,11 @@ class CommandView extends React.Component{
 
   componentDidMount() {
     this.props.command.on('change', this.updateStateFromModel, this);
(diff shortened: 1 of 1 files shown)


pcottle:
Great! Yeah tbh this was probably just vestigial code that made sense in a backbone+jQuery world (where we had to manage the DOM much more manually) but doesn't apply to a react version.

Great to see another high quality PR from you 😇 merging!

PR #671, opened 2020-03-30
chore: production build with `tinyify`

Hongarc:
Currently, `Learn Git Branching` was built with dev mode(React Developer Tools extension let me know it)

I use `tinyify` to use the production build( x2 time to build)

Another way use `process.env.NODE_ENV = 'production';` (faster and smaller origin)

|                 | Time(s) | Size(kB) |
|-----------------|---------|----------|
| Origin          | 27      | 2179     |
| tinyify         | 57      | 1873     |
| Change NODE_ENV | 25      | 1913     |

Use `tinyify` spend too much time, if you want to use `Change NODE_ENV`, just close this pull request and I create another PR or you can add `process.env.NODE_ENV = 'production';` on `miniBuild` function.

Related http://browserify.org/#more

PR #672, opened 2020-03-31
Using the production build of React

Hongarc:
#671


pcottle:
And the environment variables get reset automatically after this? 

Thanks again for looking into all of this 🙌🏻

PR #673, opened 2020-04-04
Russian translation is supplemented for level intro2.

esinmy:
Russian translation was not complete. Now it is supplemented and has the same info as English text.


pcottle:
Awesome, thanks for the contribution! 🎉

PR #674, opened 2020-04-07
korean translation typo revised & some improvement for polite speech

jeonhyun97:
I noticed a typo in Korean translation, so I simply fixed it (line 1427)

Also, although most of the Korean translations are written in polite speech, few of them were written in informal speech, which were quite awkward. For consistency, I fixed them to polite speech.


pcottle:
Oh I wasn't aware of the polite vs informal speech, interesting to learn :D Thanks for the contribution! 🎉

PR #680, opened 2020-04-21
Optimize load js and minify css+html

### LINKED ISSUE #679: Wrong input text history
Reproduce:
 - Open website
 - Type any command, ex: `ls`
 - Press up arrow: terminal add a little space after ls

hong4rc:
Hi Peter.

This PR has many changes, tell me something that you don't like:

 - Merge all CSS to one file, minify CSS and HTML, fixed #679 (space in textarea)
 - Move `font` folder to `assets`
 - Merge `jquery-ui` to `bundle.js` (don't create too many requests)
 - Defer parsing of JavaScript: load js when parsing HTML(currently wait to parse HTML done)
   - Add `defer` to script
   - Move `bundle.js` to head(load first), but `bundle.js` call template(script with `type="text/html"`) so I moved that script to head too


package.json
```diff
@@ -36,10 +39,11 @@
   "dependencies": {
     "backbone": "^1.4.0",
     "flux": "^3.1.3",
-    "jquery": "^3.4.0",
+    "jquery": "^1.8.0",
```
pcottle, line 42: this seems like a pretty big jump backwards in jquery histrory -- anything to be concerned about?

gulpfile.js
```diff
@@ -141,8 +130,14 @@ var miniBuild = function() {
 };
 
 var style = function() {
-  return src('src/style/main.css')
-    .pipe(gHash())
+  var chain = src('src/style/*.css')
+    .pipe(concat('main.css'));
+
+  if (process.env.NODE_ENV === 'production') {
```
pcottle, line 136: ah yeah I remember we set this in the miniBuild function. its a bit less direct than the config param earlier but this works

pcottle:
Boom, should be out in prod! Seems to take a smidge longer while building now but a worthwhile tradeoff for speedier page loads 😇

hong4rc:
Backbone require jquery in node_modules to work with browserify so I add
jquery to package.json(don't care version because I don't see error).

I build this PR change with jquery 3 and failed.

pcottle:
hrm are you sure @hong4rc ? once I ran `yarn` to upgrade, I was able to get everything building just fine. its also in the prod branch right now:
https://github.com/pcottle/learnGitBranching/commits/gh-pages

and everything seems to be working great

package.json
hong4rc: > hrm are you sure @hong4rc ? once I ran `yarn` to upgrade, I was able to get everything building just fine. its also in the prod branch right now:
> [`gh-pages` (commits)](https://github.com/pcottle/learnGitBranching/commits/gh-pages)
> 
> and everything seems to be working great

:)
 You forgot call `yarn` to install `jquery v1.8`
image

gulpfile.js
hong4rc: `buildIndex` is called in `fastBuild`(`ifyBuild` don't set `process.env.NODE_ENV`) and `build`(`miniBuild` set `process.env.NODE_ENV` -> `'production'`)

package.json
waldyrious: For reference, there's further discussion about this in #681.

yarn.lock
```diff
@@ -2499,10 +2563,10 @@ jasmine@^3.1.0:
     glob "^7.1.4"
     jasmine-core "~3.5.0"
 
-jquery@^3.4.0:
-  version "3.4.1"
-  resolved "https://registry.yarnpkg.com/jquery/-/jquery-3.4.1.tgz#714f1f8d9dde4bdfa55764ba37ef214630d80ef2"
-  integrity sha512-36+AdBzCL+y6qjw5Tx7HgzeGCzC81MDDgaUP8ld2zhx58HdqXGoBd+tHdrBMiyjGQs0Hxs/MLZTu/eHNJJuWPw==
+jquery@^1.8.0:
+  version "1.12.4"
```
pcottle, line 2567: sounds like this is what broke things....

before we were truly on 1.8, and going to 1.12 had the issue with drag and drop.

the weird thing is that if I go down to 1.8 for dependencies and build, i get some odd errors.

PR #683, opened 2020-04-22
Revert jquery back to 1.8 or above

### LINKED ISSUE #682: Drag and drop for git interactive rebase doesn't work
Hi,
When trying to change the order of commits when doing for example `git rebase -i HEAD~4` (as in level Moving Work Around:2) nothing happens. Tried this in different browsers (chrome, firefox, Internet explorer) on linux and windows pcs.
And thanks for this wonderful site btw

pcottle:
Resolves #682 and #681  for reference

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
   "dependencies": {
     "backbone": "^1.4.0",
     "flux": "^3.1.3",
-    "jquery": "^3.4.0",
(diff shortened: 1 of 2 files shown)


pcottle:
the branch looks fine to me in travis, just this UI isnt updating... weird

Issue #688, opened 2020-04-23
Unable to get past "hide goal" phase at the very first tutorial


ilgazcan:
Hello,

I was unable to do anything after the "Goal To Reach" window pops up. When I click any of the window buttons of the goal window, nothing happened. Then I tried to write down `hide goal` on the console; no luck either. 

This error comes up on the browser console: `this.goalVis is undefined`

I tried this on an Ubuntu Linux, Firefox, Brave and Chrome's latest browsers. I saw the issue on every one of them

Thanks!

elodg:
Same issue here. Goal dialog cannot be closed.

shubham908:
I am also getting the same issue. It worked fine just an hour back but now it has this issue.

neilteng:
just facing the same issue

ChangMarkusYu:
Facing the same issue here. Goal dialog cannot be closed and the visuals inside the goal window aren't displaying either.

abc12345131:
Same thing here. I'm a first time user, I thought it's my problem until I saw this issue.

owleyes:
Same here. Reproduced on the following three browsers before I came here to check if others see the same issue.


- Brave Version 1.7.98 Chromium: 81.0.4044.113 (Official Build) (64-bit)
- Chrome Version 81.0.4044.122 (Official Build) (64-bit)
- Safari  Version 81.0.4044.122 (Official Build) (64-bit)

ValentiWorkLearning:
The same issue.

iantolan11:
I think adding:
  <script defer src="lib/jquery-ui-1.9.0.custom.min.js"></script>
to line 361 of src\template.index.html might fix this

shinyT:
the sme issue

Liji-i:
Just facing the same issue. I have tried chrome and edge. Neither works.

hong4rc:
> I think adding either:
> 
> <script defer src="lib/jquery-ui-1.9.0.custom.min.js"></script>
> to line 361 of src\template.index.html
> 
> or changing line 95 of gulpfile.js to:
> entries: [...glob.sync('src/**/*.js'), ...glob.sync('src/**/*.jsx'),'lib/jquery-ui-1.9.0.custom.min.js'],
> 
> should fix this

We use jquery 3.5, jquery-ui 1.9 cannot compatible with it, I create a PR to fix this issue.

Quick demo: https://hong4rc.github.io/test/index.html

[...]

PR #689, opened 2020-04-23
Update gulpfile.js

iantolan11:
fixed issue with goal not initializing

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -92,7 +92,7 @@ var buildIndex = function(done) {
 
 var getBundle = function() {
   return browserify({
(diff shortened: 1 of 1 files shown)


iantolan11:
For issue #688

hong4rc:
Merged #686, It can't fix #688, I will create a PR to fix it

pcottle:
I don't think this is needed anymore since we require jquery ui via an explicit require statement and it's packaged in the overall bundle via yarn 🤔

pcottle:
Gonna close this out since I don't think its required :)

PR #698, opened 2020-04-27
Don't  check redundancy branch to compare tree

### LINKED ISSUE #691: i want to close "hide goal", i can not see anything, bu i can not close it using cmd
image

### LINKED ISSUE #697: Can't Close Level 2: Branching in Git
Can't Close Level 2: Branching in Git

If I created a new branch with a typo in the name, and then the correct bugFix branch, the lesson does not read, I need to delete all the branches and create a new one to go through this lesson, for a beginner it is not clear.

Example: 

  git checkout –b bugFi
  git checkout –b bugFix

hong4rc:
Fixed #697 

We don't need to check all branch of `treeA.branches` (have redundancy branch)

Resolve with:

    git checkout -b bugFi
    git checkout -b bugFiX

https://hong4rc.github.io/test/index.html?NODEMO&command=level%20intro2;git%20checkout%20-b%20bugFi;git%20checkout%20-b%20bugFix (resolved)

https://learngitbranching.js.org/index.html?NODEMO&command=level%20intro2;git%20checkout%20-b%20bugFi;git%20checkout%20-b%20bugFix (unresolved)

### DIFF
--- a/src/js/graph/treeCompare.js
+++ b/src/js/graph/treeCompare.js
@@ -92,13 +92,7 @@ TreeCompare.compareAllBranchesWithinTrees = function(treeA, treeB) {
   treeA = this.convertTreeSafe(treeA);
   treeB = this.convertTreeSafe(treeB);
 
-  var allBranches = Object.assign(
-    {},
-    treeA.branches,
-    treeB.branches
-  );
-
-  return Object.keys(allBranches).every(function(branch) {
+  return Object.keys(treeB.branches).every(function(branch) {
     return this.compareBranchWithinTrees(treeA, treeB, branch);
   }.bind(this));
 };


pcottle:
Thanks for looking into this! The fix here is tricky though so let me summarize:

- The original issue is because we are comparing *all* branches of both the sandbox current state and the goal state. So an extra branch in the current state, even if the goal totally doesn't care about it, causes the level to not progress.

Now we have different comparison modes in a lot of the later levels, some of which ignore this issue since they never compare branches
https://github.com/pcottle/learnGitBranching/blob/master/src/js/graph/treeCompare.js#L49

the way the method is written is actually correct, but this happens to work since we always call this function with `treeB` being the goal tree and treeA being the state tree.

im ok with this fix but kinda want to make it clear here, since before this commit you can reverse the arguments and things always work

pcottle:
Might go with #699 if it passes tests. Its a bit overkill but I want to be clear that treeA and treeB are no longer arbitrary names after this commit but mean different things :P

pcottle:
ok going with that other version, but thanks so much @hong4rc  for tracking down the issue here! can't believe this has been a issue for years and it only came up now haha

PR #703, opened 2020-05-04
Add git to Begining of Code Example

amayer5125:
Reduce confusion if "commit" is a command or a sub-command of git.


pcottle:
Good point! Hopefully people don't have `commit` as a command but you never know :)

🎉

Issue #705, opened 2020-05-05
Missing Period


amayer5125:
There is a missing period at the end of the sentence "`git clone` will be that command" on the Our Command to create remotes modal. I haven't check if any other languages are missing periods in the same place.

Also command should probably be lower case in the modal title.

amayer5125:
There is also a missing period in the last sentence of the Git Fetch modal after "This ties into the previous lesson on remote branches". I wasn't sure if I should open a separate issue for each one or post them all here. I figured a comment was less spammy.

pcottle:
Wanna submit a PR like in #704? 😇

nem75:
Hi, I opened https://github.com/pcottle/learnGitBranching/pull/707.

It adds the requested period characters. And a couple more. ;)

(For some reason the PR wasn't linked automatically?)

amayer5125:
@nem75 GitHub [looks for keywords](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) in the body in the format:
```
Git Commit Title

Other git commit body stuff
closes #705
```
Just a heads up for future commits.

nem75:
Thanks @amayer5125! Obviously I was too lazy to look it up. :) At my work we put the issue key as prefix into the title; will adapt when comitting to github.

PR #707, opened 2020-05-12
#705 Add missing period characters

nem75:
Adds period characters wherever they have been missing in complete sentences.

(Exception: asian languages / fonts. Wasn't sure if adding 'normal' period characters would have been correct there.)


pcottle:
Thanks for this! yeah I would skip any languages we aren't sure about.

Apologies for being pretty inconsistent on this :P

pcottle:
Deploying now!

PR #708, opened 2020-05-13
Using `marked` instead of `markdown`

hong4rc:
`markdown` Unmaintained (v0.5.0 Published 7 years ago)

`marked` Unmaintained (v1.0.0 Published 22 days ago)

`marked` is faster than `markdown`
image

using `benchmark` https://repl.it/@Hongarc/marked


pcottle:
Sounds great! We even removed those two other dependencies that markdown had.

And using something last updated 22 days aoge is a lot better than 7 years ago :P

PR #715, opened 2020-06-19
Bug: reverse meaning in French

gmounie:
bugFix is the parent of master, not its child. master is "descendant"
of bugFix, not the reverse.
I use "ancêtre", as other latin langage translaton, instead.
"parent" exists in French but is ambiguous, and have broader
and narrower contextual meanings (respectively, any relative, eg. cousins,
or just father and mother).


pcottle:
Great thanks a bunch! 🎉

PR #726, opened 2020-08-10
💄 Choose proper fonts for all 3 OSes

yzhang-gh:
Thanks for the great tool.

This PR tries to add proper default fonts for Windows and Linux. The font list is based on [what VSCode chooses](https://github.com/microsoft/vscode/blob/466da1c9013c624140f6d1473b23a870abc82d44/src/vs/editor/common/config/editorOptions.ts#L3464-L3466).
(I believe it is really rare for a terminal to use `Courier` as the default font.)


pcottle:
Great!!! I'm not sure the app always works in windows/linux, but I guess with a compatible version of chrome it should.

Anyways thanks so much for the PR, will merge and ship

PR #727, opened 2020-08-11
#726 follow-up

yzhang-gh:
Thanks for merging #726 in. I checked the result (on Windows) and realized I forgot to change the font family in `main.css` 😅.


pcottle:
Ok! Yeah I was surprised no css changes were needed :P

PR #731, opened 2020-08-27
Fix: Editing an existing level was broken

jannoteelem:
As discussed [here](https://github.com/pcottle/learnGitBranching/issues/187), editing an existing level was broken, e.g. `build level intro3`, resulted in an error `Uncaught Error: need goal tree and solution`. After debugging, the problem seemed to be that editLevelJSON was never assigned and when the level editor started, the level object was empty.


pcottle:
Perfect!! thanks so much for fixing

PR #733, opened 2020-09-07
Fix a typo in Russian in branching section

paraketsova:
In original English text: 

> For now though, just remember that a branch essentially says "I want to include the work of this commit and all parent commits."

In Russian translation, it uses plural form ("of these commits"), which looks like a typo, because the branch points to a single commit (and its parents), not many commits (and its parents).

This change replaces plural form ("of these commits") with singular form ("of this commit") in Russian string.


pcottle:
Thanks for this! Will push the site

PR #734, opened 2020-09-16
Added git switch with '-' and '-c' as optional arguments

mexanga:
As discussed in #596 , git has added new "alternative" commands to `git checkout`, in particular `git switch`.

`git switch <branch>` is similar to `git checkout <branch>`.

The `-c` option has been added allowing to switch to a new branch, the result is similar to `git checkout -b <branch>`.


pcottle:
Amazing! I will try to pull this down and add some test coverage perhaps?

pcottle:
Sweet, just copied some of the checkout test coverage. pushing to the site now, thanks for the contribution :D

PR #745, opened 2020-10-14
Adding support to Mexican Spanish Lang

AndreyPootMay:
First translations using _**'es_MX'**_ for i18n, for #hacktoberfest purpose.

Any comment or review is accepted.

I know that the repository does not have hacktoberfest enabled but I would love to contribute anyway and thus finish the challenge.


pcottle:
Sure but I think we will still prioritize es_LA over es_MX?

pcottle:
I guess if you set the locale code manually it will work :)

PR #752, opened 2020-11-02
[WIP] Polish translation

olsza:
Currently, 4 first lessons translated, ie "Introduction Sequence"
I will try to translate all lessons as soon as possible.

### DIFF
--- a/src/js/dialogs/confirmShowSolution.js
+++ b/src/js/dialogs/confirmShowSolution.js
@@ -154,6 +154,16 @@ exports.dialog = {
         'Verjamem vate! Maš ti to! Ali pač ne?'
       ]
     }
+  }],
+  'pl': [{
+    type: 'ModalAlert',
+    options: {
+      markdowns: [
+        '## Czy na pewno chcesz zobaczyć rozwiązanie?',
+        '',
+        'Wierzę w Ciebie! Możesz to zrobić'
+      ]
+    }
   }]
 };
 
--- a/src/js/dialogs/levelBuilder.js
+++ b/src/js/dialogs/levelBuilder.js
@@ -305,4 +305,23 @@ exports.dialog = {
       ]
     }
   }],
+  'pl': [{
+    type: 'ModalAlert',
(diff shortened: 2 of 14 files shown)


pcottle:
This is amazing work! 👏 It still shows up as a "work in progress" though so I'll hold off on reviewing it for now?

olsza:
Currently, there are 4 tasks translated ... the effects of the translation can be seen here https://git-nauka.czlowiek.it/ these are the same translations that I will send here on "pull requests", there are also my other additions on the site which I will not send, because on the website I turned off other languages, translated the menu or other that are hard in .js files;)
Certainly, the PL translations end up here before they hit the page :)

I hope that by December I will be able to translate it :)

ps. @pcottle but if you have any comments, write, it's better to improve on a regular basis :)

pcottle:
> Certainly, the PL translations end up here before they hit the page :)

Awesome! Yeah it's totally fine to work iteratively if you want -- you could put up a PR request every time you translate a new section and I can merge it iteratively. it's probably better to do that than a huge massive merge at the end which might have conflicts

(btw there are ways to default to Polish based on the browser headers and other tweaks -- not sure what other tweaks you had in your local version. but I do agree that translating everything is not super easy 🙈)

pcottle:
Whoops I clicked on "ready for review" -- would you be ok to merge this now? Seems like all the checks have passed :)

pcottle:
should be on the production site soon!

olsza:
The improvements that I can upload to you are 3 (but it would be a separate PR)
1) Change the language to "Git Branching" instead of "English"
2) change ".helperBar" attribute in CSS "max-width" to increase by 10px or "div.helperBar.show" delete "height" attribute or change it to "auto" - To make everything fit, even exit this menu.
3) window "target" shift to the right, not one on top of the other.

these solutions are in total at my site https://git-nauka.czlowiek.it/ and these are only suggestions

ps. I can see that you have played my commits;) and I will also send you up to date :)

PR #755, opened 2020-11-06
Fix css + I in README

olsza:
I am sending a small amendment, where currently the "exit from the menu" icon is not visible in languages (I send current and corrected views)

Plus I added myself to the README;)

Actual:
actual

After the amendment:
after

PS. I restored the box width to the width before my corrections were uploaded ... that is from 520px to 500px


pcottle:
haha yep I noticed this small css change but I figured it was needed for another language. thanks for the followup!

PR #756, opened 2020-11-06
[suggestion]change link with attribute target="_ blank"

olsza:
It's just a suggestion ... the link instead of opening in the same window to interrupt the "fun", opens here in a separate window. HTML was used because in Markdown, there is no option to open in a new window (I didn't find that information at least).


pcottle:
Yeah good point! This PR has the style changes in your other PR but it'll merge fine

pcottle:
thanks for all the PR's 👏

PR #757, opened 2020-11-06
[PL]Translate new info - switch

olsza:
I gave a new Polish translation for the new switch command.

ps. I made a link in HTML tags to add target="_ blank" to open in a new browser window... propose to do it in the original EN ... I also send PR #756 (either accept or reject) ;)


pcottle:
Woohoo! First translation of the new screen :)

PR #760, opened 2020-12-06
fix uk translation

vaivankov:


src/levels/remote/lockedMaster.js
```diff
@@ -19,7 +19,7 @@ exports.level = {
     "en_US": "Locked Master",
     "de_DE": "Gesperrter Master-Branch",
     "ru_RU": "Заблокированная ветвь master",
-    "ru_RU": "Заблокована гілка master",
+    "uk": "Заблокована гілка master",
```
pcottle, line 22: hrm maybe the git diff is just rendering weird but shouldnt this be removing the `uk` entry? its showing up as if its still mapped to the russian translation

pcottle:
Oohhhhh nevermind, I clearly haven't had my morning coffee. I thought you meant UK as in united kingdom (en_UK) vs Ukrainian :)

Issue #751, opened 2020-11-01
New GitHub repos will have master replaced with main


firasm:
Hi all - great resource, thank you for building this!

Just logging a teeny-tiny nit-picky feature request for the future - new git repos will have the "master" branch name set to "main" instead. See [this](https://www.techrepublic.com/article/github-to-replace-master-with-main-starting-in-october-what-developers-need-to-know/) link for more.

I wonder how disruptive it would be for the tutorial if someone were to do a "find and replace" of all instances of `master` to `main`?

It's likely not to matter for experienced git users, but it might cause beginners to trip up a bit. I'll be warning my students.

olsza:
Unfortunately, the mere use of "find and replace" is not enough, but I will probably use this terminology in the Polish translation when I am still in the initial stage.
ps. thanks for the link with the information.

pcottle:
@firasm thanks for bringing this up! I'm actually doing a lot of biased terminology migration work over at Instagram so this is top of mind during my day job as well 😇

The search and replace probably isn't sufficient (especially in all languages) but we could probably take the following steps:
- make the `master` and `main` refs always resolve to the same logical ref, so they can be used interchangeably. this means you wont be able to have two different branches with master/main names, but it'll greatly facilitate migration
- change any branch with the above ref to display `main` instead of `master`
- search and replace all the references in the english docs to refer to `main` instead of master

Do you think that would be a good first step? It might be a bit confusing that all these commands work with `master` as well as main, but maybe I could add a warning or something

firasm:
Sorry, I missed this! Got slammed with end-of-term stuff. 

your steps above make a lot of sense to me! Let me know if I can help in any way.

pcottle:
Alrighty @firasm, I made a pretty decent first-stab at getting this through with the steps above. Check out the work below and perhaps try a few levels and let me know what you think!

The only known problem for me right now is that error messages will continue to refer to master, but this achieves a pretty good surface-layer migration (without breaking backwards compatibility)

🎉 

<img width="990" alt="Screen Shot 2020-12-27 at 11 22 59 AM" src="https://user-images.githubusercontent.com/1135007/103177249-19031b80-4836-11eb-960e-ad043096f9dc.png">
<img width="1017" alt="Screen Shot 2020-12-27 at 11 25 10 AM" src="https://user-images.githubusercontent.com/1135007/103177270-418b1580-4836-11eb-937b-bd3f9063ea10.png">

[...]

PR #765, opened 2020-12-27
make all main references master under the hood

pcottle:
Starts to address #751 -- replacing master with main


pcottle:
Think this is a solid start!

PR #767, opened 2020-12-29
Another chunk of polish translation

YourSenseiCreeper:
Moreover I added a translation for buttons
after passing a level


pcottle:
Thanks for adding those intl strings for the dialog! Didn't realize none of the other languages could translate those buttons 🤦 

And 👍 to the added pl translations 🎉

PR #772, opened 2021-01-04
Improve styles for interactive pop-up windows

waldyrious:
- The title bar of floating windows that can be dragged (e.g. the "Goal To Reach" ones) now shows a "grab" cursor when hovered, and a "grabbing" cursor when being dragged.
- The buttons in those windows, which are clickable, now show the `pointer` (hand) cursor, instead of the `default` (regular arrow) cursor.

This PR was inspired by #439 (but is, AFAIK, unrelated to it).

### DIFF
--- a/src/style/main.css
+++ b/src/style/main.css
@@ -260,6 +260,14 @@ body.hgMode .visBackgroundColor {
   min-height: 400px;
 }
 
+div.ui-draggable-handle {
+  cursor: grab;
+}
+
+div.ui-draggable-dragging div.ui-draggable-handle {
+  cursor: grabbing;
+}
+
 div.canvasTerminalHolder > div.terminal-window-holder > div.wrapper {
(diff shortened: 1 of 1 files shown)


pcottle:
Amazing! Thanks so much

pcottle:
Just tried this out, feels way better 🎉

waldyrious:
There are a few things that would improve the experience even more IMO:

- Disabling the current behavior of clicking anywhere inside draggable windows to close them
- Changing the now dated OSX-style window controls for a single close button with an X, since all three buttons (when clickable) do the same
- Making all floating windows draggable, and closeable with the window controls
  - Alternatively, removing (or at least greying out) the controls from the floating windows that can't be closed using them

These are all independent changes that could be done separately, but in aggregate I feel they'd improve the website's UX quite a bit.

PR #773, opened 2021-01-04
Update the label from totallyNotMaster to totallyNotMain

thiagobitencourt:
Updating this label to match with other updates from `master` to `main`.


pcottle:
Good catch! I only replaced instances that had clean word breaks so it missed spots like this 😅

PR #775, opened 2021-01-05
Resolves #758 - Set github-corner height in CSS style instead of HTML attribute

### LINKED ISSUE #758: Ribbon is taking up full screen after pull Request #747
I realized today that the new ribbon added in pull Request #747  is taking up the whole screen so that the actions of the commands given in the command line are no longer visible and by clicking anywhere on the screen it would take me to this site (browser used Firefox). See screenshot below. I corrected this locally by changing the height of the svg element using % instead of vmin.
image

reveliant:
Hi! 

LearnGitBranching is a really, really great tool to learn Git, and I was very disappointed with the full-page ribbon on Firefox at my work place.

I'm not sure what was the "above" proposed solution on issue #758 (removing the ribbon apart) but setting `height: 7vmin` in CSS style instead of HTML `height` attribute seems to do the job both on Firefox 68 (on work-place Ubuntu variant) and 78 (on Debian 10).

Tested successfully with preview at https://8000-dad7cab5-4dc7-4381-9974-63b28d28edbc.ws-eu03.gitpod.io

### DIFF
--- a/src/template.index.html
+++ b/src/template.index.html
@@ -445,7 +445,7 @@
   </div>
 
 <a target="_blank" rel="noreferrer noopener" href="https://github.com/pcottle/learnGitBranching" class="github-corner" aria-label="View source on GitHub">
-  <svg width, height="7vmin" viewBox="0 0 250 250" style="fill:#151513; color:#fff; position: absolute; top: 0; border: 0; right: 0;" aria-hidden="true">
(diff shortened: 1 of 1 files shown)


pcottle:
Awesome 🎊 thanks a ton for looking into this! I'll push these changes shortly so people can actually use the tool on Firefox

PR #787, opened 2021-01-31
Tamil Translation for strings

vinothmdev:


pcottle:
Awesome work! 🎉

pcottle:
Screen Shot 2021-01-31 at 9 14 48 AM
Also we get a fair amount of traffic from that part of the world -- if you want to send a link with the locale pre-configured, you can use:
http://learngitbranching.js.org/?locale=ta_IN

(as you might be already aware)

Issue #791, opened 2021-02-06
French translation still mentions 'master' branch instead of 'main'


YassinHajaj:
Hi,

In file `strings.js` looking for `^\s+fdr_FR.*master` returns two results

> 'fr_FR': 'Vous ne pouvez supprimer la branche **master**, la branche sur laquelle vous êtes, ou ce qui n\'est pas une branche',
> 'fr_FR': '<span class="fwber">Note:</span> Dans ce niveau on ne doit faire de git checkout que sur la branche **master**. Les autres branches représentées ici (entourées d\'un cadre en tirets) doivent être utilisées uniquement comme références. Comme toujours, vous pouvez cacher cette fenêtre avec "Cacher les objectifs"',

They should be replaced by

> 'fr_FR': 'Vous ne pouvez supprimer la branche **main**, la branche sur laquelle vous êtes, ou ce qui n\'est pas une branche',
> 'fr_FR': '<span class="fwber">Note:</span> Dans ce niveau on ne doit faire de git checkout que sur la branche **main**. Les autres branches représentées ici (entourées d\'un cadre en tirets) doivent être utilisées uniquement comme références. Comme toujours, vous pouvez cacher cette fenêtre avec "Cacher les objectifs"',

waldyrious:
@YassinHajaj nice catch — have you considered submitting a PR making that change? You can even [edit the file directly in the browser](https://github.com/pcottle/learnGitBranching/edit/master/src/js/intl/strings.js) :)

YassinHajaj:
Hello @waldyrious 
In fact, I tried to create an upstream branch for my changes but got a nice 403 back :D

pcottle:
Oh it's not intentional, we should switch over those references to master as well. I'm not sure how they got missed in our original pass. @YassinHajaj want to submit a PR?

YassinHajaj:
Hi @pcottle ,
I just created it: https://github.com/pcottle/learnGitBranching/pull/793
I did not rerun the app to test because it seems like these kind of changes are quite straightforward, however let me know if this is a mandatory prerequisite.

I also left some comments for you to review, which might bring us to the creation of one or multiple new issue(s)

YassinHajaj:
By the way, I put it in WIP ATM to be able to make some extra changes, cleaning up the log

pcottle:
> I did not rerun the app to test because it seems like these kind of changes are quite straightforward, however let me know if this is a mandatory prerequisite.

[...]

PR #793, opened 2021-02-07
WIP #791 - French translation still mentions 'master' branch instead of 'main' 

YassinHajaj:
I did a bit more than the French translations, taking with me some translations in other languages.

**Notes**

- Korean and Thailandese are difficult for me to change as I don't have any knowledge about the languages.
- I see Polish has a reference to 'master' in one of the translations but not in the second. This is because the word 'główna' is used instead of 'main', this might or might not be intended. I would say not.


src/js/intl/strings.js
```diff
@@ -400,23 +400,23 @@ exports.strings = {
   ///////////////////////////////////////////////////////////////////////////
   'git-error-branch': {
     '__desc__': 'One of the error messages for git',
-    'en_US': 'You can\'t delete the master branch, the branch you are on, or things that ' +
+    'en_US': 'You can\'t delete the main branch, the branch you are on, or things that ' +
```
pcottle, line 403: Ah I must have missed this entire file (but handled the level translation files), now this makes more sense :P

src/js/level/index.js
```diff
@@ -200,7 +200,7 @@ var Level = Sandbox.extend({
     var onlyMaster = TreeCompare.onlyMasterCompared(this.level);
     // first we make the goal visualization holder
     this.goalCanvasHolder = new CanvasTerminalHolder({
-      text: (onlyMaster) ? intl.str('goal-only-master') : undefined,
+      text: (onlyMaster) ? intl.str('goal-only-main') : undefined,
```
pcottle, line 203: 👍

pcottle (approved):
Wow that's awesome you inspected the other languages too! Thanks for all this great work 👏

PR #795, opened 2021-02-08
Translations for bottom command menu

YourSenseiCreeper:
Hi!

I've added new translations for bottom command menu. I've also tried to translate interactive rebase modal window, but unluckily after my changes it didn't show up. Therefore I only pushed changes I was sure about.

Have a nice day!

### DIFF
--- a/src/js/intl/strings.js
+++ b/src/js/intl/strings.js
@@ -1925,5 +1925,59 @@ exports.strings = {
     'en_US': 'Confirm',
     'pl' : 'Dalej',
     'ta_IN': 'உறுதிப்படுத்தவும்'
-  }
+  },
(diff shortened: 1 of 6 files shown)


waldyrious:
@YourSenseiCreeper nice work! While you're at it, can you also double-check the `goal-only-main` Polish translation, following @YassinHajaj's comment in #793?

YourSenseiCreeper:
@waldyrious that was a good point. Translation was wrong indeed.

pcottle:
Thanks for adding strings! I've been procrastinating about adding new entries as you can see -- I planned for some kind of codemod eventually with the `intl.todo` calls but it never happened.

Glad to merge these in! 🎉

PR #797, opened 2021-02-10
Fix text falling out in the main popup

youhymuk:
As for me, it will look better without a border for `code`, but it's up to you :wink:


pcottle:
Open to that as well! Thx for the fix :)

PR #800, opened 2021-02-20
Translate missing sentences in zh_tw.

CRTao:
This is the greatest tutorial about git branching!
I found out that some sentences are not translated in Traditional Chinese in interactive rebase dialog and command console.
Update the missing sentences.


pcottle:
Awesome thanks so much for the contribution @CRTao ! 🎉

PR #804, opened 2021-03-04
fixing the use of master inside juggling commits #1

samking314:
fixes issue #803 

However, there are still many more uses of "master" instead of "main" and they should be addressed accordingly.


pcottle:
Hrm yeah, this isn't the best solution since it only fixes this specific level.

We can do a more holistic solution by changing the `git branch` command itself. This commit is how I fixed git log:
https://github.com/pcottle/learnGitBranching/commit/9cadfa1eee05dfacda28fc4560e12dad2bfd3553

this is where we print the branches:
https://github.com/pcottle/learnGitBranching/blob/master/src/js/git/index.js#L794

instead of `branch.id` why not use resolve name? that will leverage the conversion code and should fix every level!

PR #805, opened 2021-03-04
switching use of master to main in git branch cmd

samking314:
fixes #804 and the overall issue of "git branch" command using "master" instead of "main"

*although I would note that an even better solution would be to gut the use of "master" from the repo so as to reduce confusion


pcottle:
Great! Is the `.split(")[1]` to remove the quotes? could also just do `replace` for that

thanks for the fix and congrats at becoming a contributor!

PR #809, opened 2021-03-08
Fix translations in rampup/reversingChanges

### LINKED ISSUE #808: Inconsistent task description in rampup/reversingChanges across languages
I noticed that at least in some languages the task description in rampup/reversingChanges is inconsistent. Solving the task, it is clear that one commit on `local` and one commit on `pushed` needs to be reset or reverted, respectively. The English description points that out correctly: "You will revert two commits total (one per branch)."
However, at least in some other languages, this is not clear:

de_DE:
"Um diesen Level abzuschließen musst du sowohl auf `local` also auch auf `pushed` jeweils die zwei letzten Commits zurücknehmen." 
-> This says to revert the last two commits on `local` as well as the last two commits on `pushed`.
I'd propose to reword this to:
"Um diesen Level abzuschließen musst du sowohl auf `local` also auch auf `pushed` jeweils den letzten Commit zurücknehmen."

The same issue seems to be present in gl, es_ES, es_AR and pt_BR, while it is apparently correct in fr_FR and sl_SI. Unfortunately, I cannot asses the other languages…

st31ny:
Closes #808.

### DIFF
--- a/src/levels/rampup/reversingChanges.js
+++ b/src/levels/rampup/reversingChanges.js
@@ -158,7 +158,7 @@ exports.level = {
           "type": "ModalAlert",
           "options": {
             "markdowns": [
-              "Para completar este nivel, revertí los dos commits más recientes, tanto en `local` como en `pushed`.",
+              "Para completar este nivel, revertí el commit más reciente, tanto en `local` como en `pushed`.",
               "",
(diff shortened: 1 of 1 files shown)


st31ny:
This fixes translations that were not updated in e61a712. Both fr_FR and sl_SI seam to be correct already, however, I have not checked the other languages.

pcottle:
Thanks a ton for following up! 🙏

PR #812, opened 2021-03-15
Italian translation

ShardanaSoft:
My first pull req, It tooks a bit longer cause I'had to find out how to sign back old commits.
Fix for #812 
I've build it and tests are ok. 
possible issue: When I open index.html, it automatically open in italian with learnGitBranching/index.html?**locale=it_IT**


pcottle:
Hey @ShardanaSoft! Thanks so much for putting the time into this -- seems like there are a lot of formatting changes though (tabs instead of spaces perhaps, removing the string wrappers around JSON keys, and some others) that make it hard to review.

Sorry you worked on this with 60+ commits before we caught this. Do you think you could format the files back to their original format so it's easier to review?

This way it also won't destroy the git blame history we have within the files (which I've actually used before)

ShardanaSoft:
well.... that's a way to learn I guess.
I've used vs code, and auto indentation is (was) my best friend, guess it's the one to blame about space and tabs. Gotta pay attention about string wrappers, always automatic actions.
Most of the work is done, I have just to copy paste my translations.

my plans are:
start back from master,
trying with a few commits just to see if I'm doing this correctly this time.

and about translation position? I'm appending ita on the last, is that good?
example

en
fr
es
de
**ita**

pcottle:
Yeah that sounds great! And translation position shouldn't matter -- its a keyed lookup (except for the menu I think where we can vaguely go by popularity)

PR #794, opened 2021-02-07
Fix image display in the goal popup

youhymuk:

PR #822, opened 2021-04-16
Revert "Fix image display in the goal popup"

### LINKED ISSUE #817: Goal Dialog Dragging
Hello!

Thanks for the wonderful tutorial!

I noticed that the Goal Dialog doesn't like to move. When I click the title bar and drag it, the dialog does nothing. However if I go to drag it again, it teleports and closes. 

 
dVNhSOLJ8W

I'm on Windows 10 and I've tried this on Edge and Chrome.

Let me know if there's any other info I can give you—or a way I can help. Thanks!

hong4rc:
Hi @pcottle

Long time no see :)

This reverts commit e110bcedd0951587801490a344e90d542a779c13.(#794)

Fixed #817

We must use another way to fix that bug (I don't know what is that bug)

### DIFF
--- a/src/style/main.css
+++ b/src/style/main.css
@@ -296,7 +296,6 @@ div.canvasTerminalHolder.startTree > div.terminal-window-holder div.inside {
 }
 
 div.canvasTerminalHolder > div.terminal-window-holder div.inside {
-  position: relative;
   min-height: 200px;
   border-radius: 0 0 5px 5px;
 }


pcottle:
I don't even remember what #794 was fixing, but I definitely agree that moving the goal window seems more important than that.

(I vaguely remember that the goal tree was rendering *outside* the window on some browsers, but that might have been another issue)

Let me check what this looks like...

pcottle:
Oh yay, totally fixed! @youhymuk  do you remember what #794 was fixing?

PR #823, opened 2021-04-16
change master to main in CLI output when checking out via click

cbrenberg:
This adds a regex to replace 'master' with 'main' in the command line output when checking out main by clicking on the tag with the branch name.

Resolves issue #821.

FYI: @pcottle - as we discussed in the issue comments, I wasn't able to build the app to test this change

### DIFF
--- a/src/js/visuals/visBranch.js
+++ b/src/js/visuals/visBranch.js
@@ -467,7 +467,12 @@ var VisBranch = VisBase.extend({
       return;
     }
 
-    var commandStr = 'git checkout ' + this.get('branch').get('id');
+    var branchName = this.get('branch').get('id');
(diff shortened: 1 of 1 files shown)


pcottle:
Ah yes this totally works! But I have an idea on how to consolidate the logic a bit. here I'll show in a followup PR

pcottle:
Yep here:
https://github.com/pcottle/learnGitBranching/commit/4148f0b846105a0887142c5ddeda877fa6573ebd

this will use the git engine's logic to do the replacement, so its a bit more centralized :)

cbrenberg:
Thanks for the update - this is a cleaner approach for sure!

PR #828, opened 2021-04-22
A little polish to Polish translation

YourSenseiCreeper:


src/js/dialogs/sandbox.js
```diff
@@ -727,56 +727,21 @@ exports.dialog = {
...
-        '[https://pcottle.github.io/learnGitBranching/?demo](https://pcottle.github.io/learnGitBranching/?demo)',
-        '',
-        'Masz dość tego okna? Dodaj `?NODEMO` do adresu URL, aby się go pozbyć, w ten sposób:',
-        '',
-        '[https://pcottle.github.io/learnGitBranching/?NODEMO](?NODEMO)'
-      ]
-    }
-  }, {
-    type: 'ModalAlert',
-    options: {
```
pcottle, line 751: so this modal alert left?

pcottle:
Have one comment about that view that got removed but otherwise lg!

src/js/dialogs/sandbox.js
YourSenseiCreeper: I saw en_US translation without that part, so I removed it thinking it's not actual anymore.
Would you like me to bring it back?

pcottle: Oh then that's totally fine -- sorry I didn't check if it was more consistent or not.

Thanks for all your contributions on the polish translation!

PR #846, opened 2021-06-06
fix: add zh_TW translation

linooohon:
I found Locked Main level didn't have the zh_TW version, so I translated it.
Thank you for the amazing App.


pcottle:
Thanks a bunch! yeah this level is a bit more recent so it missed the wave of translations earlier

PR #863, opened 2021-07-26
Fix typo in exercise instructions for Mixed - level 5

amelielecoz:


pcottle:
Awesome! If you find more of these, mind sending them as just one PR? will be easier to merge and deploy that way 😇

PR #864, opened 2021-07-27
Tracking step tamil translation 

vinothmdev:


pcottle:
Great! Really nice to see Tamil fleshed out

PR #870, opened 2021-08-06
Disable 'git switch' along with 'git checkout'

SPGoding:
This PR disables `git switch` for levels where `git checkout` is disabled.


pcottle:
Good catch!! Totally missed this when adding the command/alias

PR #871, opened 2021-08-08
Polish translation to push and fetchRebase chapters

YourSenseiCreeper:
Hi,

There's another chunk of Polish translation. 
I've found that remote branch on fetchRebase level does not exist. Can you fix it?

### DIFF
--- a/src/levels/remote/fetchRebase.js
+++ b/src/levels/remote/fetchRebase.js
@@ -17,7 +17,8 @@ exports.level = {
     "uk"   : "Розбіжності в історії",
     "ko"   : "엇갈린 히스토리",
     "vi"   : "Dị biệt lịch sử",
-    "sl_SI": "Razdeljena Zgodovina"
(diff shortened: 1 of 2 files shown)


pcottle:
Awesome! Thanks for the translation progress 👏 

> I've found that remote branch on fetchRebase level does not exist. Can you fix it?

What do you mean? are you saying the level has some bad data or that the string isnt translated in any files? can you add a screenshot?

YourSenseiCreeper:
> Awesome! Thanks for the translation progress 👏
> 
> > I've found that remote branch on fetchRebase level does not exist. Can you fix it?
> 
> What do you mean? are you saying the level has some bad data or that the string isnt translated in any files? can you add a screenshot?

Sorry, I assumed that remote tree should be visible too. To me 'git clone' command was a little bit misguiding. Now I see because of I need to learn more about git.

PR #873, opened 2021-08-12
Minor formatting/typo fix

Qusijue:


pcottle:
Ah thanks -- yeah I can be pretty rough around the edges when formatting these exceptions (since hopefully im never seeing them in console!)

PR #875, opened 2021-08-13
rebase onto implemented

Qusijue:


pcottle:
Wow exciting you took this on! Would you like to add some tests like in these files:
https://github.com/pcottle/learnGitBranching/blob/main/__tests__/git.spec.js#L170

or should I? If you call `debug_copyTree()` in the JS console while using the app, you can get the JSON that represents the current state of the tree (which is how I wrote all those tests)

__tests__/git.spec.js
```diff
@@ -423,4 +423,27 @@ describe('Git', function() {
...
+      );
+    });
+
+    it('rebase onto fast forward', function() {
+      return expectTreeAsync(
+        'git switch -c F1; git commit; git switch -c F1; git commit;',
```
pcottle, line 444: this doesnt seem to have a `rebase --onto` call?

pcottle:
Thanks for adding tests though! Let me know if you want to update that one or we can just land as-is

Qusijue:
the test has been fixed

pcottle:
👏 👏 👏

pcottle:
(updating `gh-pages` now with your changes!)

PR #893, opened 2021-10-14
Fix main main reference

pcottle:


waldyrious:
For context, this was pointed out by @ElLutzo [here](https://github.com/pcottle/learnGitBranching/commit/b7ac2851c6fbd5a878f3db6acb84b4c5f67d9a32#r58031409).

PR #924, opened 2022-02-12
Se mejoran y añaden varios textos al idioma español de México

guepardo190889:


pcottle:
¡Muchas gracias!

src/js/intl/strings.js
```diff
@@ -1710,7 +1711,7 @@ exports.strings = {
     'zh_CN': '您正在关卡中，这里有多种形式的帮助，请选择 "help level" (关卡帮助)或 "help general" (一般帮助)',
     'zh_TW': '您正在進行關卡中，這裡有多種不同型式的幫助，請選擇 "help level" 來了解這個關卡，或者是選擇 "help general" 來學習如何使用 Learn GitBranching，或者是選擇 "objective" 來學習如何解決這個關卡',
     'es_AR': 'Estás en un nivel, por lo que hay varios tipos de ayuda. Por favor elegí entre "help level" para aprender algo más sobre esta lección, "help general" para ayuda sobre el uso de Learn GitBranching, o "objective" para aprender a resolver este nivel.',
-    'es_MX': 'Estás en un nivel, por lo que hay varios tipos de ayuda. Por favor elige entre "help level" para aprender algo más sobre ésta lección, "help general" para ayuda sobre el uso de Learn GitBranching, o "objective" para aprender a resolver este nivel.',
+    'es_MX': 'Estás en un nivel, por lo que hay varios tipos de ayuda. Por favor elige entre "help level" para aprender algo más sobre ésta lección, "help general" para ayuda sobre el uso de Learn GitBranching, o "bjective" para aprender a resolver este nivel.',
```
waldyrious, line 1714: I don't understand this change. "objective" is a command in LearnGitBranching, so it can't be edited.

src/levels/intro/merging.js
```diff
@@ -344,9 +344,9 @@ exports.level = {
           "type": "GitDemonstrationView",
           "options": {
             "beforeMarkdowns": [
-              "Acá tenemos dos ramas, y cada una tiene un commit que le es único. Esto significa que ninguna de las ramas incluye \"todo el trabajo\" que hay en nuestro repositorio. Hagamos un merge para solucionar eso.",
+              "Acá tenemos dos ramas y cada una tiene un commit único. Esto significa que ninguna de las ramas incluye \"todo el trabajo\" que hay en nuestro repositorio. Hagamos una mezcla para solucionar eso.",
               "",
-              "Vamos a `merge`ar la rama `bugFix` a `main`."
+              "Vamos a `merge`(mezclar) la rama `bugFix` a `main`."
```
waldyrious, line 349: There should be a space before the parenthesis.

src/levels/rampup/interactiveRebase.js
```diff
@@ -10,6 +10,7 @@ exports.level = {
     "en_US": "you can use either branches or relative refs (HEAD~) to specify the rebase target",
     "es_AR": "podés usar tanto ramas como referencias relativas (HEAD~) para especificar el objetivo del rebase",
     "es_ES": "puedes usar tanto ramas como referencias relativas (HEAD~) para especificar el objetivo del rebase",
+    "es_MX": "puedes usar tanto ramas como referencias relativas (HEAD~) para especificar el objetivo del rebase",
```
waldyrious, line 13: @pcottle does the translation system support language fallbacks? For example, ex_MX could have es_ES as a fallback, pt_BR could have a pt_PT fallback (or vice-versa) and in cases like this where the strings are identical, one could leave the string missing and let the system pick up the fallback string.

There are several advantages to such a system:
- We would avoid duplicating the same content in multiple translations
- It would be less likely that related translations would go out of sync
- Changes to the fallback language would be automatically picked up by the dependent languages
- The effort to add a new translation for language variants would be reduced
- Users would see missing strings as a language they're likely to understand better, rather than English

waldyrious: For the record, I would gladly contribute a pt_PT translation if I could lean on the existing pt_BR translation and only add the strings that would actually be different.

pcottle:
Yep there's a fallback locale here @waldyrious:

https://github.com/pcottle/learnGitBranching/blob/main/src/js/intl/index.js#L9

how should we do the fallbacks for the following locales?
- es_AR
- es_MX
- es_ES

additionally we could add `pt_PT` with a fallback to `pt_BR`

(the fallback map is a little weird -- we could actually do multiple directions of fallbacks since they only get looked up when a string is missing)

waldyrious:
Oh, I hadn't realized that! It's pretty cool that this is already available :)

I will leave it for the Spanish speakers to voice their opinion about the feedback system, but my instinct would be to set `es_ES` as the fallback of both `es_AR` and `es_MX`. If I'm understanding correctly, and bidirectional fallback is possible, I think it's OK to have es_ES also depend on those two, as long as there's logic to avoid infinite recursion.

It may be useful to look at the [fallback chain graph](https://www.mediawiki.org/wiki/File:MediaWiki_fallback_chains.svg) of the MediaWiki locales.

As for `pt_PT`, yeah I think it makes sense to add it with `pt_BR` as a fallback. From the top of my head, the main change in the text would be the use of "rama", which isn't widespread in Portugal (we just use "branch" directly).

PR #925, opened 2022-02-19
fix: update import module `marked`

hong4rc:
[Bump marked from 1.0.0 to 4.0.10](https://github.com/pcottle/learnGitBranching/commit/1c63af937752a03a591b391997a63737b73bac32)

This commit change version `marked` but does not update code

image

By the way, I update `marked` version


pcottle:
Hrm so if we merge this the backbone views will break?

hong4rc:
> Hrm so if we merge this the backbone views will break?

This image is the current status (you can install the module and build to see this error).

`require('marked@1.0.0')` is a `marked` function 
`require('marked@4.0.10')` is a object include `marked` function

So I call `marked` property of `require('marked@4.0.10')`

pcottle:
Great, sorry I was confused (doing this all from my phone while remote)

PR #927, opened 2022-02-22
fix syntax language in remote/lockedMain

NTAK666:
I'm sorry for my format, is my default by IDE
Maybe my pull request make your website has some error here :(

image


pcottle:
We can open up this PR again if you want! that breakage was not related to your changes (and should be fixed in main now)

PR #933, opened 2022-02-24
build: build for netify

hong4rc:
You should register Netlify and config like https://github.com/pcottle/learnGitBranching/discussions/932

You can config domain https://learngitbranching.js.org/ for Netlify instead of Github page or make `gh-pages` branch is a sub-folder (https://stackoverflow.com/a/36782614/6250402)
image

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1,5 +1,9 @@
 var { execSync } = require('child_process');
-var { writeFileSync, readdirSync, readFileSync } = require('fs');
+var {
+  writeFileSync, readdirSync, readFileSync,
+  existsSync, statSync, mkdirSync, copyFileSync,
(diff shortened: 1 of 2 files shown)


pcottle:
Awesome, I just set everything up for build previews. Let's start with that (and hopefully that will help catch breakages like the `marked` dependency update) and then maybe if I get brave enough, I'll have `gh-pages` deploy automatically.

My hesitation there is that sometimes I like updating the production version less often than merging PRs (so I don't have to worry about breakage as much)

pcottle:
Woohoo!

#934
Screen Shot 2022-02-24 at 8 14 02 AM

https://deploy-preview-934--xenodochial-hugle-b9ec84.netlify.app/

PR #936, opened 2022-02-25
fix: autocomplete show right shadow content and support multiple command

hong4rc:
Typing: `git   <multi-space here>     comm` in terminal:
image
image

Expected:
image

Support multiple command (split by `;`):
image

Currently, press  <kbd>tab</kbd> to replace the shadow content terminal, it must be not empty.


pcottle:
Looks way better, thanks for catching this (hadn't thought of multiple spaces). love the netlify preview too

PR #938, opened 2022-02-26
fix: show ball in bottom window

hong4rc:
Currently:
image

I am trying to show full ball:

image

Import `jquery-ui` via:
```js
require('jquery-ui/ui/widget');
require('jquery-ui/ui/scroll-parent');
require('jquery-ui/ui/data');
require('jquery-ui/ui/widgets/mouse');
require('jquery-ui/ui/ie');
require('jquery-ui/ui/widgets/sortable');
require('jquery-ui/ui/plugin');
require('jquery-ui/ui/safe-active-element');
require('jquery-ui/ui/safe-blur');
require('jquery-ui/ui/widgets/draggable');
```

so `lib/jquery-ui-1.9.0.custom.min.js` is unused, just remove it


pcottle:
Ah it does look a lot nicer when the balls bounce on the bottom of the screen! Thanks for the polish fix 🙌🏻

PR #942, opened 2022-03-03
Changed sandbox.js，with a fault description of URL

LetMeFly666:
If in China, the url would be: [https://learngitbranching.js.org/?locale=zh_CN](https://learngitbranching.js.org/?locale=zh_CN)

And if you want to get into the sandbox immediately, the right way is to add "&NODEMO=" to the end of the url ([https://learngitbranching.js.org/?locale=zh_CN&NODEMO=](https://learngitbranching.js.org/?locale=zh_CN&NODEMO=)) .

But if you add "?NODEMO" as the hint, the url would be [https://learngitbranching.js.org/?locale=zh_CN?NODEMO](https://learngitbranching.js.org/?locale=zh_CN?NODEMO). Then "zh_CN?NODEMO" would be consideredd as the locale and the language will not be in Chinese and the page will not begin with sandbox immediately.

Img

---

在中国访问该网站，url将会是：[https://learngitbranching.js.org/?locale=zh_CN](https://learngitbranching.js.org/?locale=zh_CN)

如果想要直接进入沙盒(sandbox)
，正确的方式是在url后面加上“&NODEMO=”([https://learngitbranching.js.org/?locale=zh_CN&NODEMO=](https://learngitbranching.js.org/?locale=zh_CN&NODEMO=))。

但是如果像提示中的那样直接在url后面加上“?NODEMO”，url将会变成 [https://learngitbranching.js.org/?locale=zh_CN?NODEMO](https://learngitbranching.js.org/?locale=zh_CN?NODEMO)。这样所在地locale将会被系统理解为 “zh_CN?NODEMO”
，网页将会不再显示中文，并且不会直接进入沙盒(sandbox)。


LetMeFly666:
And if we add "&NODEMO=" after the url, the language will be right and it will get into sandbox immediately.

image

并且如果我们直接在url后添加“&NODEMO=”的话，语音将会正确，并且网页将会自动进入沙盒(sandbox)。

Tisfy:
Ohh, that is.

But you just modified 1 place in js and if you are in other countries except Chana, the hint will also be wrong.

pcottle:
Actually I think I fixed `zh_CN` as well but please inspect!

pcottle:
https://learngitbranching.js.org/?locale=es_ES
https://learngitbranching.js.org/?locale=zh_CN

LetMeFly666:
> Great catch! But yeah as @Tisfy said, this only fixes it for one locale of the many we have.
> 
> We really should just be saying to add `NODEMO` to the url, and folks will have to unfortunately understand the nuances of url param encoding (aka add a `?` if its the first param, or add `&` if its another param)
> 
> I put up #943 to try to address this in as many locales as possible. For zh_CN could you help by changing the translation to specify the full URL? Then folks can just click on it and its explicit (like the other translations)

Sorry to bother you again!

But actually I think we should add `&NODEMO` but not `NODEMO` to the url.

`https://learngitbranching.js.org/?locale=es_ES` + `NODEMO` is `https://learngitbranching.js.org/?locale=es_ESNODEMO` while `https://learngitbranching.js.org/?locale=es_ES` + `&NODEMO` is `https://learngitbranching.js.org/?locale=es_ES&NODEMO` 

I don't know whether what I did is right, but need I to change the js again?

pcottle:
@LetMeFly666 I guess my point is that the sentence is still valid -- you are adding the query parameter `NODEMO` to the current URL. It's the *way* in which you add the query parameter that has to conform to URL encoding -- that means if there are no parameters you have to add `?` and if there are you have to add `&`.

Understanding that is difficult, which is why it's much simpler to just *give* them a link with it already done (which now we are doing for every locale).

I'd also point out that ideally our automatic language detection would work without adding any url parameters -- if we did that eventually, then the rules would change (we would add `?` instead of `&` even in other languages).

Hence I would rather drop all the implementation details of url encoding and just go with our current content for now if that's ok. But open to feedback here!

LetMeFly666:
Oh, I understand bro.

Sorry to bother you!

waldyrious:
> ideally our automatic language detection would work without adding any url parameters

Actually, as someone who lives in Portugal and uses a Portuguese keyboard layout, but sets up the browser and operating system to use English, I prefer the explicitness of URL parameters as opposed to software and websites automatically guessing what language I would want to use them in. The same could be said for people who travel a lot, or for some reason defy the common heuristics for auto-language detection. Not a deal-breaker, of course, but it's something to consider :)

PR #943, opened 2022-03-03
Resolves #942 -- fix demo URLs in other locales

pcottle:


pcottle:
Oh wait this has a bug, justasecond

PR #945, opened 2022-03-04
Changed `NODEMO` to `&NODEMO` in all locales

LetMeFly666:
The `&NODEMO` could be considered as a string, and could be added to the url immediately. #942


pcottle:
Discussed in #942

Issue #960, opened 2022-04-28
Accessibility issues


zcorpan:
Hi!

I tried using macOS VoiceOver to navigate https://learngitbranching.js.org/ and noticed some accessibility issues:

- The "Select a level" dialog is always present as seen by the screen reader (I guess it's positioned off-screen, should be `display: none` or so when not visible)
- Level links are not links.
- Next/previous buttons are not buttons (not focusable and no text, so impossible to reach with VoiceOver).
- The `textarea` doesn't have a label.
- Content outside modal dialog boxes are not inert and so can be reached with screen readers. `<dialog>.showModal()` could help (or maybe the `inert` attribute).
- No landmarks (e.g. `helperBarMount` could be a `<nav>`).
- The visual representation as SVG is navigable with VoiceOver and it can access the text, but it's probably still not understandable for blind users. This one might be tricky to solve, though!

For people with low vision or color vision deficiency:

- Some text have poor color contrast, e.g. after solving a level, the "Confirm" and "Cancel" buttons.
- Some text is small, e.g. "`You can hide this window with "hide goal"`".

ulyssear:
Some PRs have been created with some points solved (i guess) :

>  Content outside modal dialog boxes are not inert and so can be reached with screen readers. <dialog>.showModal() could help (or maybe the inert attribute).

Done with #1022 

>  After typing a command (e.g. git status), the resulting text that appears in the terminal should be read (it can be a live region, e.g. <output>). Compare with how VoiceOver works in the macOS Terminal app.
>  After typing a command, if the graph changes (e.g. git commit), the change should be announced (with a live region, or maybe as additional text in the terminal output).

Done with #1023. I haven't tried VoiceOver (or equivalent) yet.

waldyrious:
> Done with #1023.

For reference, this was reverted e11d691480 in following the report in #1026.

ulyssear:
> > Done with #1023.
> 
> For reference, this was reverted [e11d691](https://github.com/pcottle/learnGitBranching/commit/e11d69148086411b4ab3e030512d569d0a262a1b) in following the report in #1026. The approach may need to be refined.

Since the issue has been solved, do we put the <output> elements back or use the aria attributes instead?

pcottle:
I think aria labels would be a lot safer in terms of browser support

ulyssear:
>  Level links are not links.

Done with #1030

shushiro:
[...]

PR #961, opened 2022-04-28
Accessibility: use buttons

zcorpan:
Part of #960.

---

This is WIP in that only a single button was fixed so far (the first OK button).

With this approach, there should be no visual difference. The button is now keyboard focusable and accessible with screen readers. With macOS VoiceOver, when navigating to the button, it says "OK, button".

### DIFF
--- a/src/style/main.css
+++ b/src/style/main.css
@@ -62,6 +62,31 @@ div.modalTerminal p {
   text-align: center;
 }
 
+.visuallyHidden {
+  clip: rect(0 0 0 0);
+  clip-path: inset(50%);
+  height: 1px;
+  overflow: hidden;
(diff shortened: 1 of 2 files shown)


pcottle:
Screen Shot 2022-04-28 at 4 17 29 PM
yeah looks great!

zcorpan:
Marking ready for review. Still not all buttons are real buttons, but I might not be able to go through everything in a timely manner. Maybe the remaining can be done in a follow-up PR?

pcottle:
Screen Shot 2022-04-28 at 5 11 46 PM
checked out interactive rebase too and it seems to work :) thx for kicking this all off!

PR #963, opened 2022-04-30
add git gc command

### LINKED ISSUE #957: `git gc` implementation
It would be cool if LGB had a little tiny garbage-collector. Nothing fancy with packs, just to remove all of the commits not pointed to by a branch. I could *probably* implement this.

thetayloredman:
Little attempt at doing #957.

Closes #957.

This probably needs an animation as if it was traversing from HEAD down.

### DIFF
--- a/src/js/git/commands.js
+++ b/src/js/git/commands.js
@@ -208,6 +208,14 @@ var commandConfig = {
     }
   },
 
+  gc: {
+    displayName: 'gc',
+    regex: /^git +gc($|\s)/,
+    execute: function(engine, command) {
+      engine.pruneTree(false);
+    }
+  },
+
   pull: {
     regex: /^git +pull($|\s)/,
     options: [
--- a/src/js/git/index.js
+++ b/src/js/git/index.js
@@ -1836,7 +1836,7 @@ GitEngine.prototype.pruneTreeAndPlay = function() {
     this.animationFactory.playRefreshAnimationSlow(this.gitVisuals);
 };
 
-GitEngine.prototype.pruneTree = function() {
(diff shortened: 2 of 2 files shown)


ilikestohack (approved):

pcottle:
I forgot I had the prune tree command, so convenient!

pcottle:
checked this out locally, will push the site now :)

thetayloredman:
How would we manage to do an animation or is it too obscure to need one?

pcottle:
The removed commits don't do any kind of animation but the tree does gracefully animate to its new layout, so it looks pretty decent.

heres a demo:
https://learngitbranching.js.org/?NODEMO&command=git%20commit;%20git%20branch%20-f%20main%20C1;%20gc;%20git%20gc

thetayloredman:
Yeah. Would it be better to somehow animate the traversal from every ref in the repository (marking every commit as needed)? Also, please make sure you test it with tags

pcottle:
uh oh you're right, tags break it:
https://learngitbranching.js.org/?NODEMO&command=git%20commit;git%20tag%20foo;%20git%20branch%20-f%20main%20C1;%20gc;%20git%20gc

I think animating each commit is unnecessary, we already visually show which ones are eligible for pruning by modifying the transparency

pcottle:
tags fixed in that commit ^^

thetayloredman:
@pcottle Isn't the prune tree thing used for converting from Git to Hg, and if so, will that break the Hg switch?

pcottle:
It's a little undefined for what to do with tags in mercurial mode. But before the fix above, there was a floating tag that had the underlying commit deleted so that's probably more broken than after this fix.

PR #965, opened 2022-05-19
Remove Phabricator example link as it is no longer supported

EridianAlpha:
Phabricator is no longer being maintained and new hosted instances may not be created since June 1, 2021: https://admin.phacility.com/phame/post/view/11/phacility_is_winding_down_operations/


pcottle:
Good call -- we still use an internal fork of phabricator at Meta (which I love) but I agree it's not very popular out in the wild :D

EridianAlpha:
Thanks @pcottle! That was my first ever open source contribution, having just learnt enough git through this tutorial to be able to help :) Really appreciate your work!

PR #966, opened 2022-05-21
Fix Gitpod Link

fish-404:
Click Gitpod link can't find the file,  `master` had changed to `main`

image


pcottle:
Good catch! I thought GitHub would alias these two but I realize that doesn't work for those types of links probably. Thanks for the fix!

PR #968, opened 2022-05-25
Add: Remote branches are on local, not on the remote repository

ooker777:
See #967

### DIFF
--- a/src/levels/remote/remoteBranches.js
+++ b/src/levels/remote/remoteBranches.js
@@ -55,7 +55,9 @@ exports.level = {
               "",
               "Remote branches reflect the _state_ of remote repositories (since you last talked to those remote repositories). They help you understand the difference between your local work and what work is public -- a critical step to take before sharing your work with others.",
               "",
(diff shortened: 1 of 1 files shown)


pcottle:
Screen Shot 2022-05-25 at 8 50 12 AM
looks great!

waldyrious:
Shouldn't these be called "remote-tracking branches"? Otherwise, the nomenclature itself is confusing, given the usual meaning of "remote", and given the usage of "remote repositories" in this text with precisely that meaning.

pcottle:
That's certainly more clear naming, but its common shorthand to just refer to these as "remote branches." See here:
https://git-scm.com/book/en/v2/Git-Branching-Remote-Branches

> Just like the branch name “master” does not have any special meaning in Git, neither does “origin”. While “master” is the default name for a starting branch when you run git init which is the only reason it’s widely used, “origin” is the default name for a remote when you run git clone. If you run git clone -o booyah instead, then you will have booyah/master as your default remote branch.

PR #973, opened 2022-06-12
typo fix

humit0:
replace from `fecth` to `fetch`.

Thank you for providing good material about git branch.


pcottle:
Thanks for catching this!

PR #989, opened 2022-08-02
corrected the wrong word in the Ukrainian translation

Serge044:
Hello!
During the course, I saw a mistake in the Ukrainian translation, the word 'заставить'(or 'заставити', 'заставляти') is usually used when, for example, we want to leave something in a pawnshop as a pledge or when we want to take up all the space on the windowsill with flowers, and in the context of this task, a more appropriate word would be 'змусить' - which means to force someone(or something) to do something.
I understand that this is not a critical error, but as a Ukrainian I will be happy to help correct it, besides, I am just studying Git and GitHub in more detail, and for me this pool request is a great practice ;)
p.s. Thanks for a great learning platform!


pcottle:
Thanks a bunch :D Every edit helps and I appreciate when folks take the time to put up changes like this :)

PR #997, opened 2022-08-14
Docker build

brandonmcclure:


pcottle:
looks like this is breaking the netlify preview somehow, mind looking into this? I rely on it pretty heavily to review PRs so don't want to merge this until that's in a good state :)

brandonmcclure:
I merged your main into my branch and the the build ran successfully this time. The original error message makes it seem like a flaky test could be the culprit:
```
"Uncaught exception: Error: 'expect' was used when there was no current spec, this could be because an asynchronous test timed out"
```

pcottle:
Thanks for fixing it -- yeah the tests have some asynchronous calls so unfortunately they can be flaky, but it usually passes locally for me.

I'll go ahead and merge this, but do you want to expand in the readme section about requiring powershell? Just don't want folks to get surprised by that requirement. And maybe a sentence on what advantages this has over building locally :)

thanks a bunch!

pcottle:
(also going to wait on the last checks to run)

PR #1001, opened 2022-08-20
move text about how to contribute into it's own doc

brandonmcclure:


pcottle:
I really appreciate the PR! but I think the downside here is that you can't view this information by immediately landing on https://github.com/pcottle/learnGitBranching, and have to find the (somewhat small) link to jump over to contributing.

I think having the high-level information on how to build the project and get going is useful in the readme -- maybe we could reserve the contributing doc for guidelines on how to format the pull request proposal (given that you really only see it when you start to make a PR).

Going to close for now but appreciate the effort!

Issue #1003, opened 2022-08-27
Nodes in goal over the area


fish-404:
When I use my personal notebook to visit this site, I find the top node will over the area. When I use my work computer which has bigger display, I won't meet this problem. Inspect in my notebook, find the dynamic `top` value is a negative. My notebook size is 15.6 in. 

If I use this custom style to force the top unset, everyting is ok. 

``` css
div.canvasTerminalHolder>div.terminal-window-holder div.inside svg {
    top: unset !important;
}
```

![screenshot](https://user-images.githubusercontent.com/29678177/187019517-cfd4c8cc-30dc-4a63-9594-84574d2e305d.png)

pcottle:
Oh I've seen this before, would you want to submit a pull request to fix this? I think it happens consistently on shorter windows

fish-404:
> Oh I've seen this before, would you want to submit a pull request to fix this? I think it happens consistently on shorter windows

But when I narrow shorter the browser windows in a longer display, I won't see this. I'm not sure whether it should be applied.

pcottle:
if you have that style applied all the time and it only fixes the bug and doesn't cause anything else, it's probably worth adding? we can always remove it if it causes another issue (and ill do a bit of testing on my end too)

Issue #987, opened 2022-07-28
Better position of console and goal windows


KedamaMouse:
Having the goal window overlap with the console by default is bad design, because it hides the output of any commands.
The console doesn't really need to be that tall, so I think an obvious solution would be to have the console and goal each take up half the height of the screen.
Sure you can hide the goal window, but folks generally wouldn't have to with just a bit better use of space.

danny-lin714:
Can't agree with you more! The goal window disturbs me to learn git. 
But btw this project helps me a lot. It's awesome!

fish-404:
I use custom styles to change the goal window default shown on the right side. 
``` css
@-moz-document domain("learngitbranching.js.org") {
    div.canvasTerminalHolder {
        right: 0;
        left: unset;
    }
    
    div.canvasTerminalHolder>div.terminal-window-holder {
        left: unset !important;
    }
}
```

pcottle:
Would making the goal window simply shorter work? or is it the fact that it's floating above?

I think the issue with putting it on the right is that it can visually obscure the actual current state of the git tree, which is probably more important to see than the command history

fish-404:
I'm afraid shorter goal window may not easily to view the goal tree for those complex levels.  In my computer, the goal window will cover console exactly. I often drag to the right for every level. Not sure if others display will get the same result. 

![An example screenshot to show goal](https://user-images.githubusercontent.com/29678177/187857459-687051d4-cd34-457a-9018-fecac47093ad.png)

I use the styles to make the goal window on the right automatically like the screenshot below, not sure will it works in different sizes of display. (It will cover the GitHub info).

![An example screenshot to show goal in right](https://user-images.githubusercontent.com/29678177/187857999-9a998c96-bdb7-44db-ba37-846b025459dc.png)

pcottle:
ok I'm convinced that maybe we just start with the layout you mentioned. Things are very busy at work so I'll need a few days to be able to look into this...

pcottle:
Alright I think #1007 accomplishes this -- can y'all try out the new behavior via the netlify preview?
https://deploy-preview-1007--xenodochial-hugle-b9ec84.netlify.app/

It seems to work pretty well -- I made the goal window a bit narrower which is a bit tight on remote levels with an origin, but it is nice to see the command panel :)
[...]

PR #1006, opened 2022-09-03
fix: nodes over window in short display

fish-404:
As the discussion in #1003, add a CSS rules to avoid the goal tree over the goal window in shorter display.


pcottle:
Tested it out and this looks great! I'll merge this now -- also have something in progress for the discussion about #987

PR #1007, opened 2022-09-03
Resolves #987 -- start goal on the right

### LINKED ISSUE #987: Better position of console and goal windows

pcottle:


pcottle:
alright lets try this in production...

PR #1020, opened 2022-10-07
Traduccion lenguaje MX

Dmast00:


pcottle:
Hrm these seem to be the same translations as `es_ES`, and I think we fallback to that locale here:
https://github.com/pcottle/learnGitBranching/blob/main/src/js/intl/index.js#L11

can we only add translations that are different than es_ES?

PR #1027, opened 2022-10-11
[Bugfixes] Unable to enter commands in console

ulyssear:
Applying "inert" attribute to child elements of body prevented the user to type commands in the console.
So, the code has been moved before the call of "releaseKeyboard" and "stealKeyboard".


pcottle:
thanks for the quick fix!

PR #1030, opened 2022-10-19
Accessibility: Level links are now links

ulyssear:
Part of issue #960

Replaced div elements to links with "#" as `href` value. 
I checked on my side, no problem has been created with this modification.

### DIFF
--- a/src/js/views/levelDropdownView.js
+++ b/src/js/views/levelDropdownView.js
@@ -366,8 +366,8 @@ var SeriesView = BaseView.extend({
   className: 'seriesView box flex1 vertical',
   template: _.template($('#series-view').html()),
   events: {
-    'click div.levelIcon': 'click',
-    'mouseenter div.levelIcon': 'enterIcon'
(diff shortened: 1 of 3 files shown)


pcottle:
Alright I tested as well and looks like everything is functional. I'm trying to be a bit more cautious/safe with these accessibility PRs given that some browsers didn't support the other ones, but anchor tags are pretty standard.

The only weird thing is that you can middle-click the level links now, and it opens up a new tab with just the `#`:
https://deploy-preview-1030--xenodochial-hugle-b9ec84.netlify.app/#

thoughts on that? I actually think it would be awesome if we could customize the link target to be a command that launches just that level, like so:
https://learngitbranching.js.org/?NODEMO&command=level%20intro3

pcottle:
Ok cool, could we switch to the JavaScript void then? Just so the new tab behavior isn't rough around the edges

pcottle:
works great! thanks a bunch

PR #1035, opened 2022-11-30
Add support for `git merge --squash {branch}`

gatsbimantico:
Being a popular option in Github to squash and merge, it would be interesting to add support for this option.

**Initial scenario**
After running `git checkout -b feat;git commit;git checkout main;`
```
C2 (feat)
C1 (main) *
```

**With `merge --no-ff`**
After running `git merge --no-ff feat`
```
C2 (feat) ─┐
C1 ────────┴─ C3 (main) *
```

**With `merge --squash`**
After running `git merge --squash feat`
```
C2 (feat)
C1 ────────── C3 (main) *
```
This equivalent to just `git commit -m "Squashed merge branch 'feat' into 'main'"` as the connection with the C2 node is lost, and further `git branch -D feat` will leave C2 being not connected and ready to prune. This will show how you destroy history by using squash.

### DIFF
--- a/src/js/git/commands.js
+++ b/src/js/git/commands.js
@@ -564,7 +564,7 @@ var commandConfig = {
     ],
     execute: function(engine, command) {
       var commandOptions = command.getOptionsMap();
-      var generalArgs = command.getGeneralArgs().concat(commandOptions['--no-ff'] || []);
+      var generalArgs = command.getGeneralArgs().concat(commandOptions['--no-ff'] || []).concat(commandOptions['--squash'] || []);
       command.validateArgBounds(generalArgs, 1, 1);
 
       var newCommit = engine.merge(
--- a/src/js/git/commands.js
+++ b/src/js/git/commands.js
@@ -559,7 +559,8 @@ var commandConfig = {
   merge: {
     regex: /^git +merge($|\s)/,
     options: [
-      '--no-ff'
+      '--no-ff',
+      '--squash'
     ],
     execute: function(engine, command) {
       var commandOptions = command.getOptionsMap();
@@ -568,7 +569,10 @@ var commandConfig = {
 
       var newCommit = engine.merge(
         generalArgs[0],
-        { noFF: !!commandOptions['--no-ff'] }
+        {
+          noFF: !!commandOptions['--no-ff'],
+          squash: !!commandOptions['--squash']
+        }
       );
 
       if (newCommit === undefined) {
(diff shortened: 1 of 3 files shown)


pcottle:
Wow really awesome that you put up the functionality yourself! And figured out the somewhat gross/obtuse ways we parse command options and pass them around 🥴

Thanks a bunch for the test as well!

pcottle:
(it's still in draft but ready to merge it whenever, unless you also want to add a lesson slide for it)

gatsbimantico:
Hello @pcottle 
Thank you for this app ❤️ 
The code was really easy to follow, it's very clean.

I was just manually testing it now.
While we can do `git merge --no-ff side` it seems to need two arguments for squash `git merge --squash main side`
So I'll try to fix that.

I don't feel like I can put together a lesson for it. I guess it will be an initial state with a couple of branches, and a final state with just main, so you have to use `merge --squash` to merge to main, then `branch -D some-branch` to remove references and `gc` to prune the tree. Do you want to give it a go?

src/js/git/commands.js
```diff
@@ -568,7 +569,10 @@ var commandConfig = {
```
pcottle, line 569: so the command parsing is a little weird. itll grab the first argument *after* the `--no-ff` and put it into this options map.

but for boolean flags that doesnt make sense, so we tack them on manually. so i would do

command.getGeneralArgs().concat(commandOptions['--no-ff'] || []).concat(commandOptions['--squash'] || [])

and then it might work

pcottle:
yeah I can try to add a lesson once this lands (if folks are interested)

gatsbimantico:
Thank you,
that change worked.
Now I understand why the value of the first argument seemed to not be used.

There's something funny with the tests, it might be related to the depth of the objects being compared.
Also the error messages are not helpful because the assertion is just testing for truthy.

But I think this improvement is ready to go 🎉

pcottle:
Awesome! this looks ready to go, I'll merge and push the site now

> Also the error messages are not helpful because the assertion is just testing for truthy.

Yeah unfortunately it's just checking if the tree is in the expected state; theres no helpful debug message if that assertion is incorrect :/

pcottle:
It's live!
https://learngitbranching.js.org/?NODEMO&command=gc;%20git%20checkout%20-b%20side%20C1;%20git%20commit;%20git%20switch%20main;%20git%20merge%20side%20--squash

PR #1038, opened 2022-12-07
Support new --update-refs option for rebase

volleyms:


src/js/git/commands.js
```diff
@@ -674,7 +675,8 @@ var commandConfig = {
```
pcottle, line 675: the argument parsing in this function is pretty janky (and wrong in a lot of places) but worth adding on any arguments after the boolean flags here, otherwise this would work:
```
git rebase main --update-refs
```

but this would fail:
```
git rebase --update-refs main
```

since `main` would be stored in the options map

so maybe something like

```
generalArgs = generalArgs.concat(commandOptions['-p'] || commandOptions['--preserve-merges'] || []).concat(commandOptions['--update-refs'] || [])
```

pcottle:
Looks awesome, thanks for the added functionality and the test! 🥳 

I added an inline comment to one little area, mind updating that? then we can merge!

pcottle:
This is a little out of date so closing for now but reopen if you can fix that one bug! thanks again

PR #1052, opened 2023-01-30
'Double Exclamation marks' issue solved!

ajmalwali7:
I added three exclamation marks where ever needed in the "src/js/dialogs/nextLevel.js" and "src/js/intl/strings.js" files.


pcottle:
haha sure, we can revert if anyone finds it too much

PR #1061, opened 2023-03-22
 Fix Chinese translation

smy1999:
We usually use `切换` instead of `检出` in Chinese. Anyway, in the most other translations, it uses properly.


pcottle:
Thanks for the fix!

PR #1065, opened 2023-04-10
Translations - fix typo in zh_TW

neoalienson:
should be 反映 (reflect) instead of 反應 (respond) 
Other sentences uses 反映 properly,
https://github.com/pcottle/learnGitBranching//blob/main/src/levels/remote/fetch.js#L567


pcottle:
Thanks a bunch! will merge this and push the site, once the checks pass above

Issue #937, opened 2022-02-26
learngitbranching how to install


deniskonovalov64:
Hello, can you tell me how to properly install this program on your own computer. Thanks in advance.

pcottle:
Here are the instructions for building locally:
https://github.com/pcottle/learnGitBranching#building-yourself--contributing-functionality

PR #1072, opened 2023-05-04
#937 - Solution proposal

Netz00:
Waiting for comment, all tests passed.
I will add tests for this feature after it's approval.

### DIFF
--- a/src/js/views/gitDemonstrationView.js
+++ b/src/js/views/gitDemonstrationView.js
@@ -129,17 +129,25 @@ var GitDemonstrationView = ContainedBase.extend({
   },
 
   positive: function() {
-    if (this.demonstrated || !this.hasControl) {
+    if (!this.hasControl) {
       // don't do anything if we are demonstrating, and if
       // we receive a meta nav event and we aren't listening,
       // then don't do anything either
       return;
     }
+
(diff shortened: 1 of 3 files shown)


pcottle:
Alright super close! This definitely resets the animation but I think it also prevents *advancing* to the next screen. When I tried it out using the green "right" arrow to advance through the lesson, I just got stuck in the loop of going to the reset button -> back to the original starting point, but I could never advance to the next part of the lesson.

I think this is because that `positive` method just puts us in a loop. I think we would need to do something like this:

```
onResetButtonClick: () => {
  this.demonstrated = false;
  this.takeControl();
}
```

and then just wire up the button click

Screenshot 2023-05-04 at 4 39 33 PM
Screenshot 2023-05-04 at 4 39 24 PM

Netz00:
Yeah, sorry for that. I completely forgot to check it🤦🏻‍♂️. Should be fixed with the last commit.
I am glad the rest is okay, and I am not sure if I should add some tests.

pcottle:
Looks great! Going to do two little polish things (change it to "Reset Animation" and maybe fade it out a bit) but thanks so much :D

pcottle:
Screenshot 2023-05-05 at 4 06 03 PM

PR #1074, opened 2023-05-20
Fix zh-CN translation

Sevichecc:
In Chinese, the meaning of "父" is father, rather than parent. This does not quite match the original meaning of the English word "parent". I think it would be better to use a gender-neutral term for the translation. So using the original word "parent" would be better, because “parent” is also used in its original sense in translations for Chinese (Taiwan). 

It is the same case for the word "哥们", which means "brothers" in Chinese.


pcottle:
Sounds good, thanks for the pull request to fix this!

Issue #1077, opened 2023-06-24
Issue: Show Solution confirmation should not be asked if a user has already solved the level


SoumyaBhattacharjee:
Not sure why, but even if I solved the level and want to know the desired solution, the confirmation dialog box opens. <br/>
If I have not solved the level it is absolutely fine asking my confirmation, but if I have actually solved the level there is no point in that. <br/>
I have gone through the code and figured out how to solve it, will raise a PR. Please mention if this approach is not expected.

SoumyaBhattacharjee:
PR: https://github.com/pcottle/learnGitBranching/pull/1078

PR #1078, opened 2023-06-24
The confirmation dialog should not open after a user has solved the level (Issue: #1077)

SoumyaBhattacharjee:
Checking if the level is solved from the LevelStore <br/>
If the level is solved then don't open the dialog box <br/>
Also, stop the final animation of solving if Show Solution is called


pcottle:
Great point! It's definitely silly to ask if you've already solved the level. I'll merge this and push the sit enow

Issue #1068, opened 2023-04-22
Level: Interactive Rebase Intro, Solution doesn't match the goal


CulhaciLucas:
Rebasing from a commit hash that doesn't match the expected goal still counts the level as completed.

![commitHash](https://user-images.githubusercontent.com/126324040/233763853-16e6eddd-c95c-41e7-a76d-57e23cc15483.png)

pcottle:
Discussing in #1079

PR #1079, opened 2023-06-25
Remove compareOnlyMainHashAgnostic flag from Interactive Rebase level

SoumyaBhattacharjee:
This is regarding Issue #1068 <br/>
For interactive rebase level compareOnlyMainHashAgnostic flag was set to true, removed this flag such that it compares other branches too


pcottle:
This is actually intentional, and why we have that little disclaimer about only `main` being checked:
Screenshot 2023-06-27 at 3 20 58 PM

Most people solve the level with `git rebase -i HEAD~4` which doesn't rebase the C1 commit, which means that other branch will stay where it is. but at the end of the day, where that branch ends up doesn't really matter -- just what the sequence of commits are on `main`.

is there another way to make this more clear? or I can also remove the `overHere` branch as well

pcottle:
The reason why the overHere branch exists is for this shorthand:

`git rebase -i overHere`

SoumyaBhattacharjee:
Okay, I got the point about the overHere branch, but the other concern still is that the commit hash of c1 is getting changed and we were not allowing that in some other levels. For example, I have tried the same thing for level 1 and it is not getting passed 
LearnGitBranchingLevel1Cropped

pcottle:
That's what the "hash agnostic" part of the flag means. We don't care about if C1 got amended slightly, just the final order of the commits.

SoumyaBhattacharjee:
Okay got it, will close this then. Thanks for explaining

PR #1081, opened 2023-07-07
Update a french message to make it more natural

Mokona:
The proposed current sentence is a direct translation of the English sentence. Understandable if you know English, but not natural French.


pcottle:
Thanks for the fix!

pcottle:
(just pushed an update with these changes)

PR #1086, opened 2023-07-15
Added delete flag to 'push' command to delete remote branches

birdalicious:
Adds the delete flag as a method to delete remote branches
`git push origin -d foo`
uses the existing method of deleting remote branches by colon refspec


pcottle:
Wow and with tests too!! 👏🏻👏🏻👏🏻 thanks so much for the contribution

PR #1087, opened 2023-07-18
updated undo command to only track when a command changes the tree

birdalicious:
undo now only undo changing commands for example:
```
git commit (C1)
git commit (C2)
git log
git log
undo
```
will now revert the state of the tree back to C1
where as before each undo would correspond to a git log command first so you would need 3x undos.

This should have no affect on undos in levels as they override the sandbox behaviour and have their own logic for tracking if a command counts for golf (which seems to be the same condition only captured explicitly for each command)

### DIFF
--- a/src/js/sandbox/index.js
+++ b/src/js/sandbox/index.js
@@ -67,7 +67,8 @@ var Sandbox = Backbone.View.extend({
 
   initGitShim: function(options) {
     this.gitShim = new GitShim({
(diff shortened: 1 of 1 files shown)


pcottle:
👏 🥳🥳🥳🥳

Great idea and UX improvement! Thanks a ton for the pull request -- definitely agree this is much better than having to undo "past" commands that didn't change the git state.

Will push the site now

Issue #977, opened 2022-06-30
git rebase error log history


LiuShuaiQ:
Hello, I have a error order history when I use the last tag in main tab.

> Here we have main that is a few commits ahead of branches one two and three. For whatever reason, we need to update these three other branches with modified versions of the last few commits on main.

Blow is my git command:
```sheel
$ git rebase C2 three
$ git checkout one
$ git cherry-pick C4 C3 C2
$ git checkout two
$ git cherry-pick C5
$ git rebase two one
```

<img width="1791" alt="WechatIMG71754" src="https://user-images.githubusercontent.com/12734053/176578695-52077d6e-73a3-4d37-bccb-87f411759c86.png">

The history of branch `one` should be:
```
C2``->C3``->C4``->C5`->C1->C0
```

pcottle:
to repro:
https://learngitbranching.js.org/?NODEMO&command=level%20advanced3;%20git%20rebase%20C2%20three;%20git%20checkout%20one;%20git%20cherry-pick%20C4%20C3%20C2;%20git%20checkout%20two;%20git%20cherry-pick%20C5;%20git%20rebase%20two%20one

I *think* this is because we have some code that tries to understand the order of the commits when rebasing, and that unfortunately assumes the chronological order (what you see above) vs the other in the rebase target

I dont think it applies to most lessons because most of the time we arent rebasing out of order

LiuShuaiQ:
Ok, I don't think this is a general problem, the rebase order of other courses is normal. 

Thansks for your reply.

PR #1090, opened 2023-07-23
Rebase update

birdalicious:
Fixes issue #977, the breadth first search already has the commits in the order of tree hierarchy just in reverse.

Added the test case from the issue


pcottle:
Great catch! Yeah if all the tests pass then I'm supportive of this -- seems like it solves the bug without affecting anything else

PR #1104, opened 2023-09-19
Add switch options

kazzna:
Added '--create', '-C', and '--force-create' options.

Please let me know if this is wrong or I'm missing something to do.

### DIFF
--- a/src/js/git/commands.js
+++ b/src/js/git/commands.js
@@ -925,6 +925,7 @@ var commandConfig = {
       '-c',
       '--create',
       '-C',
+      '--force-create',
       '-'
     ],
     execute: function(engine, command) {
@@ -947,11 +948,12 @@ var commandConfig = {
       }
 
       {
-        let fc = '-C';
-        let fcOption = commandOptions[fc];
+        let sfc = '-C';
+        let lfc = '--force-create';
+        let fcOption = commandOptions[sfc] ? commandOptions[sfc] : commandOptions[lfc];
         if (fcOption) {
           let args = fcOption.concat(generalArgs);
-          command.twoArgsImpliedHead(args, fc);
+          command.twoArgsImpliedHead(args, sfc);
 
           let validId = engine.validateBranchName(args[0]);
           engine.forceBranch(validId, args[1]);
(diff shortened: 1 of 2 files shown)


src/js/git/commands.js
```diff
@@ -923,23 +923,43 @@ var commandConfig = {
...
       var generalArgs = command.getGeneralArgs();
       var commandOptions = command.getOptionsMap();
 
-      var args = null;
-      if (commandOptions['-c']) {
-        // the user is really trying to just make a
-        // branch and then switch to it. so first:
-        args = commandOptions['-c'].concat(generalArgs);
-        command.twoArgsImpliedHead(args, '-c');
+      {
```
pcottle, line 935: is there a reason for this open bracket? sorry maybe this is some new JS syntax I am unfamiliar with?

is it just scoping the `createOption` variable?

pcottle:
This is great, so much for expanding the functionality and even adding tests!!

I just had a quick question on some of the syntax. Also for my own understanding, what is a case where you would want to use force create over just create?

ah Bard answered for me:
Screenshot 2023-09-21 at 9 14 21 AM

pcottle:
pushed the site now! And removed the indenting / bracketing if thats ok. let me know if there was a reason for that :)

src/js/git/commands.js
kazzna: > is it just scoping the createOption variable?

Yes. It is just for scoping.  
I think the outer scope here would be growing, so I want to keep these temporary variables short lived.

It is completely opinion based thing, so I don't care if you want to remove them.

kazzna: That's also why I changed some variables `var` to `let`. `let` variables are removed when they go out of brackets, but `var` ones are still alive. 

https://developer.mozilla.org/en-US/docs/Glossary/Scope

pcottle: Ah I learned something today then 😇 thanks for the explanation. At Meta we don't really do this kind of sub function scoping so this is the first I've seen of it.

I started this project over ten years ago and sadly let and const weren't in the language yet, so that's why everything is var. A ton of this code could be a lot cleaner if it was rewritten today...
M

PR #1107, opened 2023-10-06
Update grabbingOneCommit.js

by-Lazarev:
There are no branch `master`, it is called `main`. SO in russian language it should be not 'in master' but 'in main branch'


pcottle:
Ah yes, we went through and renamed all `master` instances to `main` but it wouldnt catch the translated version (like `мастер`)

PR #1112, opened 2023-11-12
pt_BR Translations fix and add.

maxoliverbr:
Fixed and added some missing pt_BR translations.

### DIFF
--- a/src/js/intl/strings.js
+++ b/src/js/intl/strings.js
@@ -1756,7 +1756,7 @@ exports.strings = {
     'es_AR': 'Estás en el constructor de niveles, por lo que hay varios tipos de ayuda. Elegí entre "help general" para ayuda sobre Learn GitBranching y "help builder" para ayuda sobre el constructor de niveles',
     'es_MX': 'Estás en el constructor de niveles, por lo que hay varios tipos de ayuda. Elige "help general" para ayuda sobre Learn GitBranching y "help builder" para ayuda sobre el constructor de niveles',
     'es_ES': 'Estás en un constructor de niveles, por lo que hay varios tipos de ayuda. Elige "help general" para ayuda sobre Learn GitBranching o "help builder" para ayuda sobre el constructor de niveles',
-    'pt_BR': 'Você está no construtor de nívels, então há vários tipos de ajuda. Selecione "help general" ou "help builder"',
+    'pt_BR': 'Você está no construtor de níveis, então há vários tipos de ajuda. Selecione "help general" ou "help builder"',
     'gl'   : 'Estás no constructor de niveis, polo que hai varios tipos de axuda. Elixe "help general" ou "help builder"',
     'fr_FR': 'Vous êtes dans l\'éditeur de niveaux, donc plusieurs formes d\'aide sont disponibles. Merci de sélectionner soit "help general" soit "help builder"',
     'ru_RU': 'При создании уровней доступны несколько видов помощи. Выбери между "help general" и "help builder"',
@@ -2061,8 +2061,7 @@ exports.strings = {
     'vi': 'Hộp thoại hoặc văn bản này chưa được dịch ra ngôn ngữ của bạn rồi :( Hãy lên GitHub để cung cấp bản dịch!',
     'sl_SI': 'Ta dialog ali tekst še ni preveden za tvoj jezik :( Skoči na github in dodaj prevod!',
     'pl' : 'Kolejne okno dialogowe lub tekst nie zostało jeszcze przetłumaczone na język polski :( Zapraszam na GitHub, aby pomóc w tłumaczeniu!',
-    'it_IT':
-      "Questo messaggio o testo non è ancora stato tradotto nella tua lingua :(  Fai un salto su GitHub e dacci una mano!",
+    'it_IT': "Questo messaggio o testo non è ancora stato tradotto nella tua lingua :(  Fai un salto su GitHub e dacci una mano!",
     'ta_IN': 'இந்த உரையாடல் அல்லது உரை உங்கள் தமிழில் இன்னும் மொழிபெயர்க்கப்படவில்லை :( github-இன் வழியாக் நீங்களும் மொழி பெயர்க்க உதவலாம்!',
     'tr_TR': 'Bu ileti veya metin henüz dilimize çevrilmemiş :( Çeviriye yardımcı olmak için GitHub\'da bize katılın!'
   },
@@ -2142,6 +2141,7 @@ exports.strings = {
     'de_DE': 'Zurücksetzen',
     'zh_TW': '重新開始',
     'it_IT': "Reset",
+    'pt_BR': 'Resetar',
     'pl' : 'Wyczyść'
   },
   ///////////////////////////////////////////////////////////////////////////
--- a/src/js/intl/strings.js
+++ b/src/js/intl/strings.js
@@ -1540,7 +1540,7 @@ exports.strings = {
     'pl'   : 'Rozwiązany!!!\n:D',
     'it_IT': "Risolto!!!\n:D",
     'ta_IN': 'தீர்க்கப்பட்டது!!!\n:D',
-    'pt_BR': 'Resolvido!!!\n:D",
+    'pt_BR': 'Resolvido!!!\n:D',
     'tr_TR': 'Tebrikler!!!\n:D'
   },
   ///////////////////////////////////////////////////////////////////////////


maxoliverbr:
Fixed a type and last minute missing translation in pt_BR.

src/js/intl/strings.js
```diff
@@ -293,7 +292,7 @@ exports.strings = {
     'es_AR': 'Otro commit más, y van...',
     'es_MX': 'Otro commit más, y van...',
     'es_ES': 'Hagamos un commit. Atlanta, Georgia! Escucha WHATUPRG también.',
-    'pt_BR': 'Commitando.. Vai Timão!',
+    'pt_BR': 'Commitando.. Vai Fluzão !',
```
waldyrious, line 295: I don't understand the motivation for this change. I'm Portuguese, not Brazilian,, but still, I would be able to guess the meaning of Timão, whereas for Fluzão I had to google it (it refers to the Fluminense team in particular, apparently?). I assume there are Brazilians who are not into football who may also not be familiar with "fluzão". And in any case, I'm not sure referring to a specific team (as opposed to a generic term for any team) would be desirable here.

All that said, I am definitely out of the loop here, so feel free to enlighten me :)

src/js/intl/strings.js
```diff
@@ -359,7 +358,7 @@ exports.strings = {
...
     'es_ES': 'El repositorio remoto se ha desviado del repositorio local. Subir cambios no es un avance en cámara rápida (por eso fue rechazado tu push). Por favor haz pull para descargar los cambios en el repositorio remoto para que los incorpores en la rama actual. Los cambios se pueden hacer pull con el comando "git pull" o "git pull --rebase"',
-    'pt_BR': 'O repositório remoto divergiu do repositório local, então enviar suas mudanças não é um simples fast forward (e por isso seu push foi rejeitado). Por favor, faça pull das novas mudanças do repositório remoto, incorpore-os a este ramo, e tente novamente. Você pode fazê-lo com git pull ou git pull --rebase',
+    'pt_BR': 'O repositório remoto divergiu do repositório local, então enviar suas mudanças não são um simples fast forward (e por isso seu push foi rejeitado). Por favor, faça pull das novas mudanças do repositório remoto, incorpore-os a este ramo, e tente novamente. Você pode fazê-lo com git pull ou git pull --rebase',
```
waldyrious, line 361: I disagree with this change. The verb "é" was referring to "enviar", not to "mudanças", so the singular form was the grammatically correct choice.

src/js/intl/strings.js
```diff
@@ -713,7 +712,7 @@ exports.strings = {
     'es_AR': 'No podés hacer reset en el modo detached. Usá checkout si querés moverte',
     'es_MX': 'No puedes hacer reset en el modo separado. Usa checkout si quieres moverte',
     'es_ES': 'No puedes hacer reset en el modo separado. Usa checkout si quieres moverte.',
-    'pt_BR': 'Não se pode fazer reset no modo detached. Use checkout se quiser se mover',
+    'pt_BR': 'Não se pode fazer reset no modo detached. Use checkout se quiser mover',
```
waldyrious, line 715: Here too I disagree. The message is referring to moving `HEAD`, which is kind of a "you are here" pointer that automatically follows you around as you move in the repository (from branch to branch, making commits, etc.). So indeed "se mover" is appropriate here.

src/js/intl/strings.js
```diff
@@ -985,6 +986,7 @@ exports.strings = {
     'sl_SI': 'Oddaljeno',
     'pl'   : 'Zdalny',
     'it_IT': "Remoto",
+    'pt_BR': 'Remote',
```
waldyrious, line 989: Não deveria ser "Remoto"?

src/js/intl/strings.js
```diff
@@ -1753,7 +1756,7 @@ exports.strings = {
     'es_AR': 'Estás en el constructor de niveles, por lo que hay varios tipos de ayuda. Elegí entre "help general" para ayuda sobre Learn GitBranching y "help builder" para ayuda sobre el constructor de niveles',
     'es_MX': 'Estás en el constructor de niveles, por lo que hay varios tipos de ayuda. Elige "help general" para ayuda sobre Learn GitBranching y "help builder" para ayuda sobre el constructor de niveles',
     'es_ES': 'Estás en un constructor de niveles, por lo que hay varios tipos de ayuda. Elige "help general" para ayuda sobre Learn GitBranching o "help builder" para ayuda sobre el constructor de niveles',
-    'pt_BR': 'Você está no construtor de nívels, então há vários tipos de ajuda. Selecione "help general" ou "help builder"',
+    'pt_BR': 'Você está no construtor de níveis, então há vários tipos de ajuda. Selecione "help general" ou "help builder"',
```
waldyrious, line 1759: Nice catch! :smiley:

src/js/intl/strings.js
maxoliverbr: Yes it is a tiny little team that never won the Libertadores and I thought it would be nice to give them this tiny recognition. :)

src/js/intl/strings.js
maxoliverbr: I'm not a professional but this sounded wrong, so I used a tool https://languagetool.org/pt-BR to check the translation and it was marked as a gramatical error: "Possível erro de concordância verbal."

src/js/intl/strings.js
maxoliverbr: This is really a style issue, maybe the best would be: "Não é possível fazer reset no modo detached. Para mover-se, use checkout." I also checked the other translations in Spanish and they didn't use "si quieres si moverte".

src/js/intl/strings.js
maxoliverbr: This one like Main I left as Remote.

src/js/intl/strings.js
maxoliverbr: 👍

pcottle:
Thanks for the extra pair of eyes in reviewing @waldyrious! Looks like this is ready to go out in its current state?

pcottle:
Went ahead and pushed the site with these changes but we can always update them later!

src/js/intl/strings.js
waldyrious: I wouldn't say it's a style issue — the verb "mover" is transitive, i.e. it requires an object; so ending with sentence with it and no object afterwards is incorrect grammar.

Here' what's being moved is one's position in the repository, hence the "se" in "mover-se". In Spanish they do the same thing, but they use the "te" in "moverte" to the same effect.

src/js/intl/strings.js
waldyrious: Yes, it is a possible error if you're uncertain about what is the subject — the tool is correct in being cautious. But as a human who knows the language, I can say for sure that the subject is "enviar", or in other words, "o envio" — which is singular. So again, it should remain singular.

src/js/intl/strings.js
waldyrious: I suppose this is something for @pcottle to decide — would you be OK to replace a generic term for "team" with a term for a specific team, in this case [Fluminense](https://en.wikipedia.org/wiki/Fluminense_FC)? IMHO it's best to stay neutral in this tool, both to avoid rubbing people the wrong way (who may have strong preferences in terms of teams) and to ensure a more broadly known term is used instead of one that requires specific football knowledge.

waldyrious:
> Looks like this is ready to go out in its current state?

> Went ahead and pushed the site with these changes but we can always update them later!

Sorry @pcottle but I think it would have been more appropriate to give me at least a few hours (!) to react before merging the changes. As you can see from my responses above, I still disagree with some of the changes that were merged. Please hold for me and @maxoliverbr to reach a consensus before applying further changes — thanks! :pray:

pcottle:
Apologies! reverted for now :P

src/js/intl/strings.js
maxoliverbr: I was just following the fun of "Go Bears!" and some of the other translations. In Brazil "Timão" has a dual meaning, it's the direct translation of "big team" but it is also known of the team Corinthians with a large number of fans.

src/js/intl/strings.js
maxoliverbr: Ok I agree.

src/js/intl/strings.js
maxoliverbr: ok I agree

src/js/intl/strings.js
waldyrious: > In Brazil "Timão" has a dual meaning, it's the direct translation of "big team" but it is also known of the team Corinthians with a large number of fans.

I had this suspicion indeed. In this case I suppose it's a coin toss to use one or the other, even if "Fluzão" is unambiguously referring to a team and the other could be interpreted as "big team" as you said. I won't object to either choice, but I'd like @pcottle to chime in (@maxoliverbr could you unresolve this thread so it's not hidden from view until a decision is made?).

waldyrious:
Unfortunately PRs can't be un-merged and re-merged, so I'm afraid either @maxoliverbr, you or I would need to submit a new PR. In any case, it seems like agreement has been reached in all the threads above — but if you could chime in on the first thread above, that'd be appreciated :)

maxoliverbr:
Hi ! Just submitted a new PR with the rolledback translations  :)

waldyrious:
Sorry both, I've been quite busy in the past few days. I just saw PR #1114, but I'm a bit confused — hadn't all of this PR's changes been reverted in #1113 already?

Hmm, it seems like the [merged diff](https://github.com/pcottle/learnGitBranching/commit/4a07d70927867c44b7deeffe62dbe34bdd672755) of #1114 is indeed empty; it's a little puzzling that the diff in the PR actually showed changed content, though.

In any case, it seems like the remaining changes from this PR (minus the reversals done in #1114) still need to be re-submitted as a new PR, right?

pcottle:
@waldyrious ah yes it looks like you're right -- `git-dummy-msg` is still the string before this pull request (`Vai Timão!`) rather than the string after.

there were definitely changes in #1114:
https://github.com/pcottle/learnGitBranching/pull/1114/files

but it wasn't a complete revert.

Let me know how we should proceed -- we can revert-my-revert if necessary 😅

waldyrious:
No worries, I'll resubmit those changes as a new PR to make things easier :) I'll retain @maxoliverbr's authorship of the commit, of course.

waldyrious:
Ok, done: #1119.

PR #1117, opened 2023-11-20
Update strings.js

maxoliverbr:
Some more PT-BR translations.

PR #1119, opened 2023-11-27
Update pt_BR translations (replay of #1112)

waldyrious:
Replays changes from #1112 that were reverted in #1113 and not reintroduced in neither #1114 nor #1117.

I'm not really sure what happened there but at least this way the changes from that first PR don't get lost :)


pcottle:
Thanks a bunch!

PR #1122, opened 2024-01-07
Translate the fakeTeamwork lesson to es_MX

acHeMatta:
Translated the remote/fakeTeamwork lesson to es_MX


pcottle:
Thanks a bunch! Neat to see the differences between `es_EX` and `es_ES` in the body of the lesson :)

PR #1123, opened 2024-01-11
Translate command-not-supported to es_AR, es_MX and es_ES

chistrobal93:
There is no error message for git-error-command-not-supported, a common error that occurs when typing any command incorrectly.

- **Before fix** (missing translation message)
image

- **After fix**
image


mgarciaisaia (approved):
👍 LGTM!

¡Gracias! 😄

pcottle:
Thanks a bunch!

pcottle:
(also its neat how each one is slightly different -- really shows the difference for each locale)

PR #1126, opened 2024-02-06
Added vitejs as a launcher for index.html. package.json 

Angra974:
Hi.
Here is some little modification i've done as i don't like to load file with index.html directly ( it will open my text editor more than my browser).

Added vitejs to the package.
Add "prepare" script to package.json to call "gulp fastBuild" when installing the application first time with yarn/pnpm installation.
This will create the build folder and the index.html, file which will be used by vitejs.

Added the script "dev" in the package.json file so vite can be loaded and use index.html file to run the appliccation in our browser.

Change assets folder to /public/assets for convenience with vitejs and modified gulpfile accordingly.

Tests are done and passes without error.
image


Angra974:
> Thanks for dropping by and adding some vitejs support!
> 
> However I think the build is failing -- presumably due to this:
> 
> > 3:47:53 AM: error vite@5.0.12: The engine "node" is incompatible with this module. Expected version "^18.0.0 || >=20.0.0". Got "16.20.2"
> > 3:47:53 AM: error Found incompatible module.
> 
> seems like maybe either vite needs to be downgraded or we have to upgrade node?

Didn't think about it. I'm use to always update my package to the last version that I forgot to check the requirement of the project. I will go check for a version compatible with node16.

pcottle:
Great -- sorry for the difficulty with the build pipeline but its helped me a lot to merge PRs over time safely (even if its running on an outdated version)

Angra974:
> Great -- sorry for the difficulty with the build pipeline but its helped me a lot to merge PRs over time safely (even if its running on an outdated version)

No problem. It's your requirement, so we need to follow them.

Currently, i have no problem running it on v16.20.0 

image
image
image

I will make sone push while downgrading the version of vitejs till it will compile here without issue.

gulpfile.js
```diff
@@ -112,7 +112,7 @@ var buildIndex = function(done) {
 
   if (process.env.CI) {
     writeFileSync('build/index.html', outputIndex);
-    copyRecursiveSync('public/assets', 'build/assets');
+    copyRecursiveSync('assets', 'build/assets');
```
pcottle, line 115: do we still need this?

pcottle:
```
0.478 error LearnGitBranching@0.8.0: The engine "node" is incompatible with this module. Expected version "16.20.2". Got "14.20.0"
0.486 error Found incompatible module.
0.486 info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
-
```

gah, getting closer :/

gulpfile.js
Angra974: Yes I've got a little suggestion from vite at a time telling it will be using files from "assets" instead of public/assets so i reverted back to the original folder.

Angra974:
> ```
> 0.478 error LearnGitBranching@0.8.0: The engine "node" is incompatible with this module. Expected version "16.20.2". Got "14.20.0"
> 0.486 error Found incompatible module.
> 0.486 info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
> -
> ```
> 
> gah, getting closer :/
Tell me exactly which version of node you need for the setup.

image

I was using v20.11.0, who was i think compatible with "18.00 || >= 20.0.0" and i've added/used a 16.20.2 after that for the correction.

I've taken off the engine requirement in the package. json so it will not be checked during  the build tests with google actions.

pcottle:
Alrighty then, all the checks passed so this is good to go! Thanks for iterating on it @Angra974 :)

PR #1140, opened 2024-04-24
Resolves #1139: can use a direct permalink to a specific level

### LINKED ISSUE #1139: Add direct link to each level
Hello,

Thanks for this amazing work :)

I'm writing some articles on how to start with Git, and I'm willing to point readers to this awesome learning tool.

To keep it simple for readers, I would like **to point them directly to one specific level**, instead of the "home page" with all levels, but [I can't find a permalink for that purpose](https://github.com/pcottle/learnGitBranching/blob/caf480d9562918975c8f4d8417561c80012782a8/src/js/app/index.js#L192-L267).

I've tried to use the `command` query arg for that, it's working... almost. With a link like `https://learngitbranching.js.org/?NODEMO=&command=level intro1` this works, but the problem is the space in the value `level intro1` which may be replaced by `level+intro1` in some browsers or situations, and then lead to an invalid command :cry: 

To implement that, I was thinking of something like adding a new `level` query parameter, which value can be the same as the `level` command (`intro1`, `rampup3`, etc.): eg `https://learngitbranching.js.org/?level=advanced2`

Another possible fix I see would be to "unescape" the + sign when the `command` query arg is read, and the + being replaced by a single space. But this may have some unwanted side effects.

What do you think about this ?

pomeh:
This resolves the issue #1139 I just created, about permalink to levels.

I did not add unit tests as I didn't find anywhere query arguments are tested. But I've manually tested my changes through Gitpod (which is simply amazing !).

I also was thinking about a "share level" feature, but:
- this another feature
- this is out of scope of the initial issue
- this can have many implementations possible, and different UI design too
- this idea would need your validation first, before any change

Feel free to let me know what you think about all this.

### DIFF
--- a/src/js/app/index.js
+++ b/src/js/app/index.js
@@ -237,6 +237,10 @@ var initDemo = function(sandbox) {
       'levels'
     ];
     commands = commands.join(';#').split('#'); // hax
+  } else if (params.hasOwnProperty('level')) {
+    commands = [
+      'level ' + unescape(params.level),
+    ];
   } else if (params.gist_level_id) {
     $.ajax({
       url: 'https://api.github.com/gists/' + params.gist_level_id,


src/js/app/index.js
```diff
@@ -237,6 +237,10 @@ var initDemo = function(sandbox) {
       'levels'
     ];
     commands = commands.join(';#').split('#'); // hax
+  } else if (params.hasOwnProperty('level')) {
+    commands = [
+      'level ' + unescape(params.level),
```
pomeh, line 242: I've not implemented checks on the `params.level` format/value here.

I'm not really confortable with this idea, as [an app must always validate input args/user values before using them](https://owasp.org/Top10/A03_2021-Injection/), but I've not implemented it myself to stick to the coding style for others parameters, where there is no check either.

Issue #1138, opened 2024-04-18
Incorrect implementation for git fetch/pull with colon ":" affecting remoteAdvance6


juanmv94:
In _remoteAdvance6_ level we execute the command **git fetch origin foo~1:bar** to move remote _foo~_ into local _bar_ branch, but that's actually not allowed by Git, resulting in a **fatal: Invalid refspec 'foo~1:bar'** message.

When using colon ":", origins can be a remote branch name, or commit hash, no ~^ modifiers allowed. Also when using a remote branch as origin, example: **git fetch origin foo:bar** it will update _bar_ but also updates _o/foo_ not being reflected in the simulator.

These same issues affects **git pull** commands in the same way

pcottle:
Good catch! Do you think just switching to commit hashes instead would solve the inconsistency but still provide a valuable lesson?

juanmv94:
Definitely that could be a workaround until it’s fixed on the simulator, although I think the most common way of using the colon fetch is with a branch name origin (updating o/branch)

pcottle:
Alright @juanmv94 just updated the level -- let me know what you think:
https://learngitbranching.js.org/?NODEMO&command=level%20remoteAdvanced6

juanmv94:
**More issues discovered affecting fetch and pull commands:**
I'm making a pull request fixing all these issues including the previous ones

### _git fetch origin foo:bar_ (with HEAD on o/foo up to date, bar branch still doesn't exists) --> Already up to date!

**Examples:**

git clone
git fetch origin main:foo

### _git fetch origin foo:bar_ (with current HEAD not part of history of foo) --> Your origin branch is out of sync with the remote branch and fetch cannot be performed

**Examples:**

git checkout -b foo c0
git commit
git clone
git fetch origin main:bar

git clone
git fetch origin c0:foo

PR #1141, opened 2024-04-25
Fixed many fetch/pull issues on #1138

juanmv94:


juanmv94:
Need to make some little fix to my changes...

juanmv94:
Looks like all compilation issues are related to tests expecting different (incorrect) results that were addressed by the fixes. Tests might need corrections:

* **doesn't fetch if out of sync, but will update explicit dest if specified**
o/main gets updated after _git fetch origin main:main_ but wasn't updated before correction

* **correctly resolves source during git fetch with params**
**correctly resolves existing commits and updates**
**doesn't let you fetch to main if you are checked out there**
Does unallowed _git fetch origin foo^:blah_

* **pulls to a new branch and then merges in that branch**
o/main gets updated after  _git pull origin main:bar_ but wasn't updated before correction

pcottle:
Yeah I totally agree that we shouldn't *teach* the functionality here that doesnt work (hence why I redid the lesson), but I hesitate to actually break the functionality.

There might be others relying upon this for levels in gists or with hardcoded commands, and adding the validation code might break that :/

will think on this but it seems like a lower priority item compared to updating the lesson (which we did)

pcottle:
Alrighty, looks like everything is passing! Appreciate all the updates here and the work @juanmv94 :)

Will merge and push

PR #1142, opened 2024-04-28
Implemented --force for fetch/pull. Force using +src:dest

juanmv94:


pcottle:
Awesome! Do you have a good command that would demonstrate the force functionality? was going to add a test...

pcottle:
@juanmv94 I added a basic test case for this in 6cbd02ec01f6ff8430f9b15bb401c50c98c3c3d2, here's how I did it:

- Opened up the sandbox and ran a few commands that exercised the new code
- Opened up the JS console and ran this command to copy the tree string
- Added it to the spec file
- 
Screenshot 2024-04-28 at 9 09 44 AM

Can you think of a set of commands that would only work with git pull `--force` and maybe the `+src` syntax?

pcottle:
Oh and that command pops up a dialog where you can copy the string easily:
Screenshot 2024-04-28 at 9 13 13 AM

PR #1146, opened 2024-05-13
make modal draggable

hiroto22:
This chage allows you to move the modal that comes up when doing an interactive rebase.
I think this is useful when you want to work on a modal while looking at the goal. (ex. level2 move2)

- before
スクリーンショット 2024-05-13 22 08 53

- after
スクリーンショット 2024-05-13 22 09 39

### DIFF
--- a/src/js/views/index.js
+++ b/src/js/views/index.js
@@ -327,6 +327,14 @@ var ModalTerminal = ContainedBase.extend({
     };
 
     this.render();
+
+    this.$terminal = this.$el.find('.terminal-window-holder').first();
+    this.$terminal.draggable({
+      cursor: 'move',
+      handle: '.toolbar',
+      containment: '#interfaceWrapper',
+      scroll: false
+    });
   },
 
   updateTitle: function(/*string*/ title) {


pcottle:
Awesome contribution! But I think this makes *all* terminal windows draggable, including the window for selecting a level and a bunch others.

Could we just do this change on `./js/views/rebaseView.js` so its contained to just the interactive rebase one?

hiroto22:
Thanks for the review!
I think it would be fine if all terminal windows are draggable, such as to see Instractions while checking the current conditions.

pcottle:
Hrm I wasn't able to get it to drag all the way down -- could you?
Screenshot 2024-05-13 at 7 37 40 PM

the goal window successfully can drag all the way down

hiroto60:
I wasn't able to do it.
I can do it in the local environment, but not after deployment.😕

hiroto22:
I wasn't able to do it.
I can do it in the local environment, but not after deployment.😕

pcottle:
Yeah the draggable jQuery plugin is quite old and I wouldn't be surprised if it's buggy. I might revert this change just so we avoid any bugs but appreciate the PR anyways!

PR #1170, opened 2024-09-08
Fixing the Japanese description text

hosokawat:
Fixing the Japanese description text
Markdown italic not being applied due to missing spaces

スクリーンショット 2024-09-08 12 56 34


pcottle:
Thank you @hosokawat! Yeah this is an easy thing to miss when doing the translations, good eyes on catching it

Issue #1177, opened 2024-09-20
add a link to a clean slate state


parmentelat:
when in sandbox mode, it's possible to do a `reset; clear`
however when using the tool often this takes some time - to type and to execute

I would suggest to add a link from the `Learn Git Branching` header to go back to the main sandbox entry point and restart afresh

I actually have a possible PR to this end so if the idea feels right I can go ahead with it

pcottle:
I'd be fine with a shortcut :) you could add an shortcut here that achieves that, maybe a broom or trash can or something:
<img width="186" alt="Screenshot 2024-09-20 at 5 37 42 PM" src="https://github.com/user-attachments/assets/a8565432-d72c-47e4-9ee4-a9743097cdb7">

https://fontawesome.com/icons/broom?f=classic&s=solid

pcottle:
Speaking of which, tapping that "?" actually brings you to a screen where you can reset as well:
<img width="608" alt="Screenshot 2024-09-22 at 5 44 55 PM" src="https://github.com/user-attachments/assets/13eac84d-d59f-4650-bbef-c6cd98145d6b">

PR #1178, opened 2024-09-22
for #1177: add a <a> element on page title to support link to home

parmentelat:
hey

as a first step towards #1177 I had been playing with this approach
rationale being the home icon in the header, which at first I had taken as a hint towards going back to home page

I could maybe also add a shortcut as per your suggestion but I do not yet feel comfortable messing about in this area

let me know what you think

### DIFF
--- a/src/style/main.css
+++ b/src/style/main.css
@@ -1356,3 +1356,8 @@ div.gitDemonstrationView {
   left: 21px;
   opacity: 0.4;
 }
+
+a.home-link {
+  text-decoration: none;
+  cursor: alias;
+}
--- a/src/template.index.html
+++ b/src/template.index.html
@@ -395,11 +395,13 @@
             </div>
             <div>
               <label for="commandTextField">
-                <i aria-hidden="true" class="icon-home"></i>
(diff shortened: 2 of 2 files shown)


pcottle:
Ah interesting, just pulled this down locally to play with it:
Screenshot 2024-09-22 at 5 39 22 PM

It's a good start! But I think my hesitation is that:
- its not super clear why the header is blue now
- clicking it will immediately clear your local state, possibly wiping out any progress you might have!

so I worry about people accidentally clicking it. It's also a bit of an odd visual design paradigm to have a link thats not underlined.

Its super easy to add another icon in that bottom bar. Just edit this file and add a new entry:
https://github.com/pcottle/learnGitBranching/blob/e12ed5ac00f2ad679a360513532d85f5280ea0d6/src/js/react_views/MainHelperBarView.jsx#L70

Use the icon classname here:
https://fontawesome.com/icons/broom?f=classic&s=solid

and then execute something like this to perform the reset:
https://github.com/pcottle/learnGitBranching/blob/e12ed5ac00f2ad679a360513532d85f5280ea0d6/src/js/react_views/CommandsHelperBarView.jsx#L37-L40

PR #1186, opened 2024-10-08
try to make it compatible with bun

RazanakotoMandresy:


pcottle:
Gonna run the full CI here but seems like a beefy diff, and the early signals show that it failed :/

PR #1204, opened 2024-11-18
feature: perfection tracker for levels with best number of commands

Patolord:
This feature introduces the ability to track and record the user's best performance on each level. By storing both the completion status (solved) and whether the user achieved the best possible outcome (best), the application can provide more detailed feedback and track user improvement over time.

### DIFF
--- a/__tests__/LevelStore.spec.js
+++ b/__tests__/LevelStore.spec.js
@@ -23,12 +23,29 @@ describe('this store', function() {
 
     expect(LevelStore.isLevelSolved(firstLevel.id))
       .toEqual(false);
-    LevelActions.setLevelSolved(firstLevel.id);
+    LevelActions.setLevelSolved(firstLevel.id, false);
     expect(LevelStore.isLevelSolved(firstLevel.id))
       .toEqual(true);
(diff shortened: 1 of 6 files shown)


pcottle:
Oh cool feature! Do you have a screenshot of what the UI looks like?

My only worry is this part:
https://github.com/pcottle/learnGitBranching/blob/928577b2dc37586a0d5945d5bddd453c69c2a506/src/js/stores/LevelStore.js#L51-L57

Since we sync that map to local storage, I think this existing code would wipe out the existing solved map status.

I think we can fix that in the `  isLevelSolved: function(levelID) {` function @Patolord

pcottle:
A star would be cool!!

I added a fix for old users:
https://github.com/pcottle/learnGitBranching/commit/b3d0004427b55ec1f9250bcbb202fcc593bb0980

it works now :)
https://learngitbranching.js.org/?NODEMO&command=levels

Screenshot 2024-11-18 at 6 47 35 PM

Patolord:
its just a matter of changing react-ok-icon to react-star class on the <i> tag.
but im not sure how to implement the conditional styling.

PR #1205, opened 2024-11-19
css improvements and star icon

Patolord:
As your suggestion, added new functionality to display a star icon for "best" completed levels in the level selection interface. I tried to  maintain the existing styling and hover states while avoiding the use of modern selectors like :not() for better backwards compatibility.
image

### DIFF
--- a/src/style/main.css
+++ b/src/style/main.css
@@ -1011,8 +1011,22 @@ a.levelIcon.solved:active {
 }
 
 a.levelIcon.best {
-  border-color: gold;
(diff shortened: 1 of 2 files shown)


pcottle:
Awesome!!! Thanks for following up, this is way more fun :)

pcottle:
looks awesome:
Screenshot 2024-11-19 at 8 40 36 AM

pcottle:
It means a bunch when folks drop by and improve things :)

PR #1223, opened 2025-02-03
fix(rebasing.js): fix the order of instructions displayed in french version

Asthilde:
Move the result instructions from the beforeMarkdown to the afterMarkdown array.


pcottle:
Nice catch! Thank you so much for the contribution, really helps

PR #1232, opened 2025-03-07
Fix: Renaming locale labels

edcedcedcedc:
Standardized language names to be written in their native form (e.g., "Italiano" instead of "italiano," "italian," or "italy").


pcottle:
Good point! this makes way more sense, thanks for the update. Especially "English" being labeled as "Learn Git Branching" makes zero sense haha.

pcottle:
Pushed the site! Should update soon

PR #1239, opened 2025-04-05
Updated Demonstration Message for Intro4

rke007:
While going through the Git lessons, I found a wording issue in the message displayed during the Intro 4 level, right after rebasing onto the **`main`** branch.

The original message said:

> “Awesome! Now the work from our bugFix branch is right on top of main...”

However, since the rebase places the **`bugFix`** commits beneath the latest **`main`** commits to maintain a linear history, I’ve updated the message for clarity:

> “Awesome! Now the work from our bugFix branch is right beneath main...”

This should better reflect the actual commit structure post-rebase.

### DIFF
--- a/generatedDocs/levels.html
+++ b/generatedDocs/levels.html
@@ -1337,7 +1337,7 @@ type <code>git checkout -b [yourbranchname]</code>.</p>
 <p>Here we have two branches yet again; note that the bugFix branch is currently selected (note the asterisk)</p>
 <p>We would like to move our work from bugFix directly onto the work from main. That way it would look like these two features were developed sequentially, when in reality they were developed in parallel.</p>
 <p>Let&#39;s do that with the <code>git rebase</code> command.</p>
-<pre class="level-solution">git rebase main</pre><p>Awesome! Now the work from our bugFix branch is right on top of main and we have a nice linear sequence of commits.</p>
+<pre class="level-solution">git rebase main</pre><p>Awesome! Now the work from our bugFix branch is right beneath main and we have a nice linear sequence of commits.</p>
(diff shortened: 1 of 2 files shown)


pcottle:
ah good point, this is where the tree being "flipped" going downwards disagrees with our terminology. We use it in a few other places:

```
./levels/rampup/detachedHead.js:              "First we have to talk about \"HEAD\". HEAD is the symbolic name for the currently checked out commit -- it's essentially what commit you're working on top of.",
./levels/mixed/jugglingCommits.js:              "Here's another situation that happens quite commonly. You have some changes (`newImage`) and another set of changes (`caption`) that are related, so they are stacked on top of each other in your repository (aka one after another).",
```

also according to ChatGPT at least, its both pretty well defined and common to use
Screenshot 2025-04-05 at 2 38 11 PM
Screenshot 2025-04-05 at 2 39 13 PM

but I totally get that the commits in the visualization are actually beneath (vertically) but pointing to their parents (on top of).

pcottle:
What about something like:

> Now the work from our bugFix branch is stacked on top of main

or 

> Now the work from our bugFix branch is stacked on top of main, since it points to main. In our visualization though, its shown below main since our commit trees flow downwards

rke007:
Well, now I got it!! Thanks for the detailed explanation. I have made the changes as per your suggestion and raised a PR.

pcottle:
Awesome thanks for being open to the feedback 😄

rke007:
This is my first ever Open Source contribution 🥹 Thank you so much!

pcottle:
Amazing!! Congrats on your first contribution 😇

PR #1242, opened 2025-04-13
fix: update korean translation typo

shren207:
Replace `커다른` with `커다란`


pcottle:
Thanks for the contribution here! Really helps everyone else using the resource :)

PR #1246, opened 2025-04-26
Update rebase animation to reset HEAD

### LINKED ISSUE #1245: Rebase animation: show HEAD moving to target branch
Currently, when visualizing a `git rebase` operation, the first animation doesn't show the HEAD moving to the target branch. This can cause confusion when explaining why the "incoming changes" during conflict resolution are from the feature branch (and not from the target branch like for a merge operation).

Initial state:

```
A---B---C  main
     \
      D---E  feature *
```

When running `git rebase main`, the first thing git does is move the HEAD to the target branch (in this case `main`).

Suggested change: Update the rebase animation to first animate the HEAD moving to the target branch (before any commits are copied/replayed).

I would be happy to contribute and implement this change myself, but I would need some guidance.

I was thinking that the `GitEngine.prototype.rebaseFinish` function needs to be updated to include a new initial animation to move the HEAD to the target branch.

cuinixam:
This is an attempt to fix #1245

### DIFF
--- a/src/js/git/index.js
+++ b/src/js/git/index.js
@@ -2420,6 +2420,17 @@ GitEngine.prototype.rebaseFinish = function(
     });
   }
 
+  // Move HEAD to the commit we are rebasing onto
+  chain = chain.then(function() {
+    this.checkout(this.getCommitFromRef(targetSource));
+    return this.animationFactory.playRefreshAnimation(this.gitVisuals);
+  }.bind(this));
+
+  // add a small delay for clarity before the next animation
+  chain = chain.then(function() {
+    return this.animationFactory.getDelayedPromise(400);
+  }.bind(this));
+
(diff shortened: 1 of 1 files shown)


cuinixam:
I have tried to fix #1245 by changing the rebase animation to first checkout the base commit from the destination branch.
The problem is that at the end, the feature branch is not reset to the last rebased commit.
I need some feedback if this change makes sense.
In the meantime I will try to update the tests and try to fix the feature branch reset at the end of the rebase.

pcottle:
Hrm let me see. I think the issue is that `feature` isnt a variable we store -- we are just checked out on HEAD so when we update HEAD to the base, we lose the information that `feature` should be updated.

pcottle:
See #1247

PR #1247, opened 2025-04-26
My attempt at #1245

pcottle:
See #1246

### DIFF
--- a/src/js/git/index.js
+++ b/src/js/git/index.js
@@ -2421,7 +2421,6 @@ GitEngine.prototype.rebaseFinish = function(
     // to update this bad boy after the animation
     originalCurrentLocationToUpdate = this.getOneBeforeCommit(currentLocation);
   }
-  debugger;
 
   var toRebase = this.filterRebaseCommits(toRebaseRough, stopSet, options);
   if (!toRebase.length) {
@@ -2482,7 +2481,6 @@ GitEngine.prototype.rebaseFinish = function(
   // source to that commit
   chain = chain.then(function() {
     if (originalCurrentLocationToUpdate) {
-      console.log('in this branch');
       this.setTargetLocation(originalCurrentLocationToUpdate, base);
       this.checkout(originalCurrentLocationToUpdate);
     } else {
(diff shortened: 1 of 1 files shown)


pcottle:
@cuinixam I think this works? Let me know what you think

cuinixam:
@pcottle I've just tested it and it works. I like it! Thanks!
I was looking forward to finishing it myself today but you were much too fast for me 😀

cuinixam:
Would you be so kind and explain why my change did not work.
My assumption was that for my use case, with rebasing a feature branch, because `currentLocation` argument is not changed, the else branch will be hit and `currentLocation` will be reset to `base` and then checked out.

```
    if (this.resolveID(currentLocation).get('type') == 'commit') {
      // we referenced a commit like git rebase C2 C1, so we have
      // to manually check out C1'
      this.checkout(base);
    } else {
      // now we just need to update the rebased branch is
      this.setTargetLocation(currentLocation, base);
      this.checkout(currentLocation);
    }
```

pcottle:
@cuinixam I really appreciate you taking a stab at it! I wouldn't have done the change if you hadn't gotten it started 😄 it really helps, truly. Also if you want, you could submit a PR to add the animation delay if that feels necessary.

Your PR was quite close but I think the issue was this:

- before, we had `HEAD` pointing at `feature` which then pointed to the actual commit `E` in your example. `currentLocation` in the method is actually `HEAD` (since we didn't specify a source, it just defaults to that)
- we create all the new commits and then this line kicks in:

https://github.com/pcottle/learnGitBranching/blob/main/src/js/git/index.js#L2471-L2472

- that line is a little deceptive, but inside the implementation, it's basically saying "go get the thing right before a commit (either the branch if its there, or HEAD if not) and go update it to this newly created commit. So it takes HEAD, follows it one away to the branch `feature`, and updates where `feature` points

So in your PR when this line runs:
`this.checkout(this.getCommitFromRef(targetSource));`

it finds the target source and then checks it out... aka it assigns `HEAD` to that location! Now we lost the reference to `feature` which we need to update.

This works great for the animation but later on we never update `feature` to point at the right commit.

So my code just found a way to hold onto that feature reference (if we need to) and update it later 😅

pcottle:
(merged and pushing the site now)

PR #1249, opened 2025-05-13
Fix: Update Japanese translation

muratad:
While using the Japanese version, I encountered a sentence that felt very different from the original meaning. 
Upon investigation, I discovered it was [a Japanese meme that was created for the 2008 Beijing Olympics](https://dic.pixiv.net/a/%E9%A0%91%E5%BC%B5%E3%82%8C%E9%A0%91%E5%BC%B5%E3%82%8C%E3%81%A7%E3%81%8D%E3%82%8B%E3%81%A7%E3%81%8D%E3%82%8B). 
Since the humor from that time is no longer easily conveyed, I updated the translation to better align with the original text.


pcottle:
Wow I had no idea that little joke snuck in there! What a funny cultural relic (but I agree it is quite old). Thanks for the link explaining it :) 

Definitely will merge this and push the site now, thanks a bunch for the update

muratad:
Thanks for the quick action and for maintaining this awesome project!

PR #1250, opened 2025-05-15
Changing strings for better representation

benceBalazs:
Changing the current de_DE (german) translated strings for better representation as the currently implemented german translation feels a bit weird.

There are no code changes, only string changes - merging the branch should probably not take much time / effort.

Regards
Ben

### DIFF
--- a/src/levels/index.js
+++ b/src/levels/index.js
@@ -102,7 +102,7 @@ var sequenceInfo = exports.sequenceInfo = {
   rampup: {
     displayName: {
       'en_US': 'Ramping Up',
-      'de_DE': 'Aufstieg',
+      'de_DE': 'Eine Stufe höher',
       'ja'   : '次のレベルに進もう',
       'fr_FR': 'Montée en puissance',
       'es_AR': 'Acelerando',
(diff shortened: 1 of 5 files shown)


pcottle:
Really appreciate the updates/improvements! Thanks @benceBalazs

src/levels/index.js
```diff
@@ -149,7 +149,7 @@ var sequenceInfo = exports.sequenceInfo = {
     tab: 'remote',
     displayName: {
       'en_US': 'Push & Pull -- Git Remotes!',
-      'de_DE': 'Push & Pull -- entfernte Repositorys',
+      'de_DE': 'Push & Pull -- Remote Repositories',
```
pcottle, line 152: is it standard to use english works for these kinds of things sometimes?

benceBalazs: In most cases we (and especially people with IT relevancy) also use english words. "Entfernte" could be translated into "distant" or "deleted" which sounds a bit odd and there is not a really good word to convey the meaning of "remote" in my opinion.

Maybe someone will find a better solution but in the meantime this should probably make things clearer.

pcottle: Thanks for the explanation :)

PR #1254, opened 2025-05-30
Fixed the portuguese translation from level relativeRefs.js

gifpxto:


pcottle:
Thanks! Every bit helps :)

PR #1261, opened 2025-06-27
run normalize

pcottle:
lets see if this works...


src/levels/mixed/describe.js
```diff
@@ -375,69 +372,6 @@ exports.level = {
         }
       ]
     },
-    "es_ES": {
```
pcottle, line 378: maybe lost an es_MX translation here but probably ok...

PR #1272, opened 2025-09-10
Fix typo in Korean language string

Hoto-Cocoa:
"앉을겁니다" means "I will sit down". "후회" means "regret". "않을겁니다" is "you will not" so new translation text will be "You will not regret".

That is common typo.


pcottle:
Thank you @Hoto-Cocoa !

Issue #1273, opened 2025-09-10
Branch colors are too similar, causing confusion in "Merging in Git" tutorial


Sooamazing:
## Problem Description
In the "3: Merging in Git" tutorial level, there's a discrepancy between the tutorial text and the visual representation that causes user confusion.

- Current Issue:
   The tutorial text states: "here we see that the main branch color is blended into all the commits, but the bugFix color is not"
However, the branch colors are generated randomly and often appear too similar to distinguish visually. This makes it difficult for users to understand the concept being taught.

- Screenshots:
    <img width="659" height="476" alt="Image" src="https://github.com/user-attachments/assets/2ab5aca9-d314-4297-bbdf-af5ac672d883" />

## Expected Behavior

- main and bugFix branches should have clearly distinguishable colors
- The visual representation should match the tutorial text description
- Colors should be accessible for users with color vision deficiencies

## Root Cause
The current color generation uses completely random hues:
```js
javascriptvar randomHueString = function() {
  var hue = Math.random();
  var str = 'hsb(' + String(hue) + ',0.6,1)';
  return str;
};
```
This can result in visually similar colors that don't provide sufficient contrast.
And also blendHueStrings function does.

## Proposed Enhancement
To improve user experience and accessibility, I suggest replacing the random color generation with a fixed, WCAG AA compliant color palette for common branch names.

pcottle:
I'm supportive of this! I actually am colorblind myself.

Do you have a list of the WCAG AA compliant color palette values?

pcottle:
Check out #1274  @Sooamazing, what do you think? I am deploying the site now

pcottle:
(reopen if you think its not resolved)

Sooamazing:
@pcottle 
Thank you for your response. I don’t have a full ready-made palette, but I have been checking contrast ratios between branch colors, the background color, modal colors, and the main branch color using tools like [venngage’s accessible color palette generator](https://venngage.com/tools/accessible-color-palette-generator).

For reference, I also use [WebAIM’s Contrast Checker](https://webaim.org/resources/contrastchecker/) to test color combinations. For example, with the current background color (#5CBCFC) and modal color (#ff4c7e), the contrast ratios between these and a possible main branch color (#00FF7F) were 1.55:1 and 2.36:1, respectively. These fall short of the generally recommended ratio for WCAG AA (normally 4.5:1 or higher for normal text).

[...]

PR #1274, opened 2025-09-10
Issue #1273 -- consistent colors per branch that are accessible

pcottle:
See #1273


pcottle:
Its not always the prettiest but I think this will be a nice improvement!

PR #1278, opened 2025-09-17
translation: french mistake accent on "faites"

1somniak:
The verb `faire` (to do/make) in the present tense, 2nd person plural is:
* `vous faites` or `vous le faites` : there is no accent on the `i`.


pcottle:
thanks for the fix!!

PR #1282, opened 2025-09-30
Fix: typo in ta_IN and tr_TR localization JSON

Klosepa9p:
Minor spelling errors in the Tamil (ta_IN) and Turkish (tr_TR) localization objects within the JS file have been corrected; text outside parentheses has been moved inside parentheses.

pic

### DIFF
--- a/src/levels/rampup/cherryPick.js
+++ b/src/levels/rampup/cherryPick.js
@@ -1134,118 +1134,118 @@ exports.level = {
           }
         }
       ]
+    },
+    "ta_IN": {
+      "childViews": [
+        {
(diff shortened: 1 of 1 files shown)


pcottle:
Thanks @Klosepa9p for putting up these fixes! It looks like there's an issue with the diff though -- perhaps your editor has switched from spaces to tabs or something?

The entire file looks rewritten so its hard to see the changes. mind reverting back so we can preserve blame and I can review the diff?
Screenshot 2025-10-04 at 4 31 42 AM

pcottle:
Much better thank you!

PR #1283, opened 2025-10-03
fix: improve inaccurate German translation of commit color rules

MsdR24:
The English description is:

`Each commit turns a color that is the blended combination of all the branches that contain that commit.`

Which was inaccurately translated into:

`Jeder Merge Commit bekommt als Farbe eine Mischung aus den Farben seiner Vorgänger.`

The German translation implies that the new commit color is a combination of it's previous (parent) commit colors which is not the case. The new translation improves the intended meaning.


pcottle:
Much appreciated @MsdR24 ! 👏

PR #1284, opened 2025-10-03
fix git describe output format (underscore -> hyphen)

op3r470r:
git describe output uses hyphens (-) not underscores (_).


pcottle:
Nice catch @op3r470r ! I never use git tags or git describe personally so I totally forgot about even adding this functionality...

I'll merge this in and push the site

PR #1285, opened 2025-10-03
Fix helperbar text for Chinese

op3r470r:
"学习分支" and "學習分支" mean "Learn Branch" and should be replaced with "简体中文" and "繁體中文", respectively, to correctly reflect the language.


pcottle:
Thanks @op3r470r for this content update! 👏

PR #1314, opened 2026-01-10
Fix editorconfig comment format

hongarc:
## Summary
- Move comment to its own line instead of inline comment in `.editorconfig`

## Test plan
- [x] Verify the editorconfig still works correctly


pcottle:
Do you want to include this @hong4rc  or was it rolled into another PR?

pcottle:
oh I see #1315

PR #1326, opened 2026-01-23
remove the word "command" from the codeblock

blankRiot96:
image


pcottle:
Oh good catch! Total mistake here 🤭

pcottle:
thanks for the fix @blankRiot96

PR #1331, opened 2026-02-08
fix(gitDemonstrationView): fix the unavailable exit function during demonstration

Fronut:
This pull request enhances the `GitDemonstrationView` class in `src/js/views/gitDemonstrationView.js` to improve user navigation and resource management. The most important changes include adding keyboard navigation support, refining the exit logic to properly clean up resources, and ensuring teardown of visualizations.

Navigation improvements:

* Added support for the `esc` key to trigger an exit action by updating the `aliasMap`, allowing users to exit the demonstration using the keyboard. ( The old exit action is unavailable in gitDemonstrationView, which can be considered as a small bug )

Exit and cleanup logic:

* Refined the `exit()` method to release control, reset the view, and call the parent container's `finish()` method if available, with a fallback to closing just this view. This ensures proper cleanup and flow when exiting.
* Updated the `tearDown()` method to check for and tear down the `mainVis` visualization before calling the base class teardown, improving resource management.
* Simplified the `die()` method by removing the check for `visFinished`, ensuring the base class's `die()` is always called.

### DIFF
--- a/src/js/views/gitDemonstrationView.js
+++ b/src/js/views/gitDemonstrationView.js
@@ -77,7 +77,8 @@ class GitDemonstrationView extends ContainedBase {
       aliasMap: {
         enter: 'positive',
         right: 'positive',
-        left: 'negative'
+        left: 'negative',
+        esc: 'exit'
       },
       wait: true
     });
@@ -88,7 +89,16 @@ class GitDemonstrationView extends ContainedBase {
   }
 
   exit() {
-    alert('exittt');
+    this.releaseControl();
+    this.reset();
+
+    if (this.metaContainerView && typeof this.metaContainerView.finish === 'function') {
+      this.metaContainerView.finish();
+      return;
+    }
+
+    // Fallback: close just this view if no parent container is present
+    this.die();
   }
 
   receiveMetaNav(navView, metaContainerView) {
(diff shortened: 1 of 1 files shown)


Fronut:
Hi! This is my second contribution to this repository (and my **second-ever** pull request on GitHub).

I've fixed an issue where the exit function wasn't available during demonstrations by adding ESC key support and improving the cleanup logic. If there are any suggestions on the implementation or if I've missed something, I'd be happy to make adjustments.

Thank you for reviewing!

pcottle:
This one is a bit trickier to review since we don't have a ton of tests on the UI layer -- ill pull it down and play around with it.

pcottle:
I pulled this down locally and I'm not sure if its needed? when I go to the first lesson on the prod website and hit `left` key, I already go backwards to the previous screen?
https://learngitbranching.js.org/?NODEMO&command=level%20intro1
Screenshot 2026-02-08 at 1 40 43 PM

Fronut:
546786197-c48a0caa-c9de-4b4f-b3b6-17661e73f267
The problem is not about the left. It is about the left-up button shown in this picture. This close button is unavailable at this moment(before clicking"git commit"). The "esc" key is also unavailable. I fix this problem.

pcottle:
Oh yeah you're totally right! Definitely a bug, it must be because the demonstration view has to yield control back to the overall modal before the escape button works. I'll merge this, I never noticed it!

PR #1332, opened 2026-02-08
Fix canvas terminal unexpected closing while clicking anywhere or dragging

Fronut:
This pull request fixes a small bug of the terminal window behavior and appearance in the application. The changes ensure that the terminal window only closes when the explicit close control is used, and updates the cursor style from pointer to default because it is not clickable.

Terminal window interaction changes:

* Changed event handling in `CanvasTerminalHolder` so that only clicks on the close control (`div.controls div.close`) trigger the close action, preventing accidental closure from clicks inside the terminal window. (`src/js/views/index.js`) 
* Before this, if you click anywhere on the terminal window or attempt to drag the terminal window will cause it to close unexpectedly.

UI/UX enhancements:

* Updated the cursor style for the terminal window wrapper to `default` instead of `pointer`, reflecting that the area is not clickable for closing. (`src/style/main.css`)

### DIFF
--- a/src/style/main.css
+++ b/src/style/main.css
@@ -296,7 +296,7 @@ div.ui-draggable-dragging div.ui-draggable-handle {
 div.canvasTerminalHolder > div.terminal-window-holder > div.wrapper {
   height: 100%;
   box-shadow: 0 0 12px rgb(0,0,0);
-  cursor: pointer;
+  cursor: default;
   border-radius: 0 0 5px 5px;
 }
 


Fronut:
This is my third Pull Request in Github. I find it really interesting!

pcottle:
> Before this, if you click anywhere on the terminal window or attempt to drag the terminal window will cause it to close unexpectedly.

oh yeah this is super frustrating, great catch! will merge now, just saw the nice improvement locally :)

PR #1333, opened 2026-02-08
fix: Fix terminal autoscroll when opening Help

Fronut:
I discovered a bug in terminal.
When opening Help immediately after entering a level, the terminal prompt appeared mid-way through the help output. The top portion kept the normal terminal style while everything below used a different “selected” style. Scrolling didn’t fix it; only typing a character would jump the prompt to the bottom and unify styles.

Here is my pull request to fix this bug:
Ensure the terminal always auto-scrolls after help (or any new output) renders by invoking the scroll logic post-render and measuring layout via requestAnimationFrame
Use scrollHeight-based sizing and proper class toggling so the terminal consistently anchors the prompt at the bottom and applies the correct styling without user input.

### DIFF
--- a/src/js/react_views/CommandHistoryView.jsx
+++ b/src/js/react_views/CommandHistoryView.jsx
@@ -27,6 +27,11 @@ class CommandHistoryView extends React.Component {
     Main.getEvents().on('clearOldCommands', () => this.clearOldCommands(), this);
   }
 
+  componentDidUpdate() {
+    // Ensure we stay pinned to the bottom after any render
+    this.scrollDown();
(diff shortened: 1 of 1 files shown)


Fronut:
My fourth Pull Request, Fixing bug is really interesting......

pcottle:
That's a tricky bug! Curious if you had a screenshot of the bad behavior but this makes sense overall, will merge this PR and then push the site.

Thanks for all the contributions @Fronut !

Fronut:
image
This is a snapshot of this bug @pcottle

PR #1341, opened 2026-03-07
Improve level completion messages to reflect user performance

CodeCopAkash:
Improved wording of level completion messages to provide clearer feedback
based on user performance relative to the optimal solution.

The dialog now reflects whether the user:
- matched the optimal solution
- exceeded the optimal solution
- used more commands than the optimal solution

No functional changes were made.

Before :

Level1matched

Level2exceeded

Level3worse

After:

Level1matchedImproved

Level2exceededImproved

Level3worseImproved


pcottle:
Thanks so much @CodeCopAkash ! This is indeed a nice little polish on the experience; we should definitely give people extra credit when they exceed the solution.

(I noticed the yarn lock got replaced as part of this PR, I might drop that while pushing if thats ok)

Thanks so much for contributing!

Issue #1233, opened 2025-03-28
git gc (garbage collection) should reset commit indices


ohadschn:
## Issue
Suppose I start with: 
`c0 <- c1(main)`

I do `git reset c0` and `git gc` so now I have: 
`c0(main)`

Finally I do git commit and get:
`c0 <- c2(main)`

Where I would have expected:
`c0 <- c1(main)`

## Impact
Beyond the matter of general consistency, this "current commit index" hidden state behavior interferes with level building.
Suppose I want the student to simply do `git commit` on a trivial graph (with only init commit). So starting from `c0(main)` I do:
* `define start`
* `git commit`
* `define goal`

But then when I share the level and the student does `git commit` they will get `c0 <- c1(main)` (and not `c0 <- c2(main)`) so the exercise won't succeed...

This is especially troublesome considering how `build level` seems to always start with `c0 <- c1(main)` (where `reset` just brings it back to that position)

## Workaround
Reset the state to a "clean" trivial tree with:
`importTreeNow {"branches":{"main":{"target":"C0","id":"main"}},"commits":{"C0":{"parents":[],"id":"C0","rootCommit":true}},"HEAD":{"target":"main","id":"HEAD"}}`

pcottle:
> Where I would have expected:
> c0 <- c1(main)

I think this expectation is up for debate -- when you perform `git reset C0`, the C1 commit still exists. It's still on your computer/filesystem and is perfectly valid, and could even have other branches pointing to it still!

`git commit` creates a new, unique, independent change that then is formed as a new commit. Thus you don't arrive back at C1, but you arrive at C2 instead (since your changes are different). Basically `git commit` would never produce the same set of changes twice, at least in this learning environment :)

I totally agree though that its a troublesome burden while building levels -- you have to make sure to not make any mistakes (or inflate your commit hash at all).e

ohadschn:
Perhaps you missed my point about `git gc`, after which C1 no longer exists.

Issue #1226, opened 2025-02-13
rampup4 should use git revert HEAD^ instead of git revert HEAD


jeqcho:
This app is amazing.

On the tutorial to level rampup4, it might be more realistic to show `git revert HEAD^` instead of `git revert HEAD`, since `git revert HEAD` doesn't seem to have any practical effects.

![Image](https://github.com/user-attachments/assets/904e2285-82a0-4574-8a9e-50e020a1759f)

jeqcho:
My bad! I misunderstood what `git revert` does. On that note, might be great to clarify that `C2'` on the graph means undoing the changes in `C2`.

jeqcho:
Since later on the symbol `C2'` is used for cherry-picking, maybe for reverts we can use strikethroughs like` ̶C̶2̶` to distinguish as a revert instead of a copy?

![Image](https://github.com/user-attachments/assets/7e167a16-1ddf-437c-883d-c1a6215332b6)

pcottle:
It doesn’t really distinguish between a revert or any other modification—
the hash just means it’s a modified version of that commit (including
reverts).

If we did the strike through then it might be tricky to describe the revert
of the revert :) I’m not opposed to it though!

-Sent from Gmail Mobile

PR #1373, opened 2026-07-08
show solution writes into the command box, plus gc/level fixes

NmaaAlhawary:
- level: `show solution` now writes the level's solution into the command box (without running it) so students can read, edit, or run it themselves, via a new commandBox_setText event + setInputText handler in the command prompt view. Split the level regexMap into levelRegexMap.js so it can be parsed/tested without loading the React views.
- git gc: rewind the commit-index counter after pruning so freed commit numbers get reused (e.g. reset + gc + commit yields C1, not C2) (#1233)
- rampup4: use `git revert HEAD^` in the revert demonstration so it has a practical effect, updated across all locales (#1226)
- help: document sandbox/level commands so `help {command}` and `show commands` cover them (e.g. importTreeNow); leave bare `git`/`git help` to the general git help
- fix: import GlobalStateStore in app/index.js; it was referenced in the beforeunload handler without an import, throwing a ReferenceError on unload
- tests: add gc-index regression test and sandbox/level command specs

### DIFF
--- a/__tests__/git.spec.js
+++ b/__tests__/git.spec.js
@@ -200,6 +200,13 @@ describe('Git', function() {
     );
   });
 
+  it('reuses freed commit indices after gc (#1233)', function() {
+    return expectTreeAsync(
+      'git reset C0; git gc; git commit',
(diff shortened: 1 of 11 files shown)


pcottle:
This all looks amazing @NmaaAlhawary ! Thanks for the very thoughtful additions/improvements. 

I was hesitant about the `show solution` change just because it requires an additional step though -- but I think it's a fine change. I thought about auto-submitting the command if you dont edit it after N seconds but thats too complicated

PR #1379, opened 2026-07-14
Real-world levels (working directory, conflicts, PRs) plus a couple of fixes

jiteeshghodke456-del:
Hi! I've been learning git with LGB and it's one of the best tools out there, so I had a go at extending it toward the everyday git people run into at work. I know dropping this much into someone else's project is a lot, so please treat it as a starting point and tell me what's worth keeping and what isn't.

New levels (all optional, graded on the commit graph, kept as beginner-friendly as the existing tracks):

- **Working Directory:** staging with `git add`/`git status`, undoing with `git restore`, and shelving work with `git stash`
- **Conflicts:** resolving a merge conflict, with an optional rebase-conflict follow-up
- **Collaboration:** the three pull-request merge strategies, and keeping a fork in sync

To teach those, I added a small opt-in "changes" model to the engine that tracks which files are staged, unstaged, or conflicted. It stays completely off for every existing level, so nothing else changes and the current tests pass as they are. I also nudged the intro to show `git switch -c` next to `git checkout`.

Two fixes I ran into along the way:

- `undo` could flip a merge commit's parents, which mirrored the tree and made `git checkout HEAD^` follow the wrong parent. Fixed with a faithful serializer and a test.
- The Korean hint on "Branch Spaghetti" was unrelated leftover text, and the Russian explanation of commit colors didn't match the English. Both corrected.

Happy to split this up, drop pieces, or rework the approach. Thanks for building such a fun tool.

Issue #1298, opened 2025-11-05
Undo Bug in Juggling Commits #1


fejfo:
I undid 2 wrong commands, and then entered the two correct ones. But the working tree still looked wrong and I had suddenly lost some commits. So I undid everything but they didn't reappear. When I entered everything again it worked.

<img width="409" height="867" alt="Image" src="https://github.com/user-attachments/assets/b56f6c9d-1a80-487c-8f10-7d059c3958cc" />

This looks like a bug with the site to me, but if I'm just wrong about git please let me know.

fejfo:
A similar bug in a later level

<img width="1915" height="713" alt="Image" src="https://github.com/user-attachments/assets/8b4ea8ce-8df4-43bf-a880-d66df71bbf96" />

Issue #613, opened 2019-10-08
`undo` caused exchange of parents


marbel82:
When I used undo, the parents of previous commit exchanged.

Bug path:
```bash
$ git commit
$ git checkout C1
$ git commit
$ git commit
$ git branch aa
$ git checkout C3
$ git commit
$ git branch bb
$ git checkout aa
$ git merge bb
$ git checkout master
$ git commit
$ git merge aa
$ git commit

$ undo
```
![image](https://user-images.githubusercontent.com/31281571/66376677-77817700-e9b0-11e9-9e0f-6828e42a0581.png)

marbel82:
State after merge:
```json
{"branches":{"master":{"target":"C8","id":"master","remoteTrackingBranchID":null},"aa":{"target":"C6","id":"aa","remoteTrackingBranchID":null},"bb":{"target":"C5","id":"bb","remoteTrackingBranchID":null}},"commits":{"C0":{"parents":[],"id":"C0","rootCommit":true},"C1":{"parents":["C0"],"id":"C1"},"C2":{"parents":["C1"],"id":"C2"},"C3":{"parents":["C1"],"id":"C3"},"C4":{"parents":["C3"],"id":"C4"},"C5":{"parents":["C3"],"id":"C5"},"C6":{"parents":["C4","C5"],"id":"C6"},"C7":{"parents":["C2"],"id":"C7"},"C8":{"parents":["C6","C7"],"id":"C8"}},"tags":{},"HEAD":{"target":"master","id":"HEAD"}}
```
It seems that `git merge` sets bad parents
```json
"C8":{"parents":["C6","C7"],"id":"C8"}}
```
but before `undo`, when I execute `git checkout HEAD^` HEAD moves to C7. It's working OK.
After undo `git checkout HEAD^` moves HEAD to C6.

pcottle:
@marbel82 this is actually a side effect of what we guarantee with `git undo`. Undo restores the state of the tree to a previous state, but does not guarantee the *ordering* of commit parents.

As you can see, this is reflected visually as the tree being flipped. This also means that commands aren't guaranteed to be able to be replayed exactly (`git checkout HEAD^` for instance uses parent ordering to know where to go).

Unfortunately at this stage its hard to fix this since the tree serialization is pretty baked in. hope thats ok!

PR #1381, opened 2026-07-15
Fix undo merge-parent bug and correct Korean/Russian dialog text

jiteeshghodke456-del:
These are the two straightforward fixes split out of #1379, as you suggested, so they can go in on their own.

- **undo bug (#1298, same as #613):** `undo` and reset snapshot the tree with a serializer that sorts merge parents alphabetically for goal comparison. Reusing it to save state meant undoing a merge flipped the parents, so the tree rendered mirrored and `git checkout HEAD^` followed the wrong parent. Fixed with a faithful serializer for snapshots; the goal-comparison path is unchanged. Includes a regression test.
- **Korean and Russian dialog text:** the Korean "Branch Spaghetti" hint was unrelated leftover text (#1150), and the Russian merge-color explanation in "Merging in Git" didn't match the English (#1296). Both corrected.

These are split out of the larger levels PR (#1379) so they can go in on their own. The new levels will come back as their own PR once they're more polished.


pcottle:
This looks great! Thanks for the PR. I'll try to push the site later today, just dont have access at the moment

pcottle:
Pushed the site!

PR #1395, opened 2026-07-30
Add ru_RU translation for the restore level

KEPASEV:


pcottle:
Thanks a bunch @KEPASEV !! It's been amazing to see all the translations come in so quickly for the new levels 👏  will push the site now

PR #1398, opened 2026-08-03
equire -d / --detach for git switch detached HEAD (#1240)

### LINKED ISSUE #1240: git switch should require -d or --detach for detached HEAD
As I said in the title. You currently don't accept -d or --detach, but git switch rejects anything that results in a detached HEAD unless -d or --detach is passed

AbdullahButt2611:
Closes #1240

Related issue: [#1240](https://github.com/pcottle/learnGitBranching/issues/1240)

### What
`git switch` now behaves like real git when the target would detach HEAD:

- Added `-d` and `--detach` to the `switch` options.
- `git switch -d <commit>` / `git switch --detach <commit>` checks out the commit and leaves HEAD detached.
- A plain `git switch <commit>` (or a tag) no longer silently detaches. It throws a git-style error pointing the user to `--detach`.
- Switching to a branch is unchanged.

### Why
As noted in #1240, real git refuses to switch onto anything that would result in a detached HEAD unless `-d` / `--detach` is passed, and it accepts those flags. The app previously accepted neither flag and detached silently.

### Changes
- `src/js/git/commands.js`: added the two options and the detach / branch-required logic in the `switch` command.
- `src/js/intl/strings.js`: added the `git-error-switch-detach` message (en_US).
- `__tests__/git.spec.js`: updated the existing "switch to a commit" test to use `-d`, and added tests for `--detach` and for the refusal case.

### Testing
- `npx jasmine "__tests__/**/*.spec.js" --random=false` passes (356 specs, 0 failures).
- No new jshint findings versus the base tree.

### DIFF
--- a/__tests__/git.spec.js
+++ b/__tests__/git.spec.js
@@ -82,11 +82,27 @@ describe('Git', function() {
   });
 
   describe('Switches', function() {
-    it("to a commit", function () {
+    it("to a commit with -d option", function () {
       return expectTreeAsync(
-        'git switch C0',
+        'git switch -d C0',
         '{"branches":{"main":{"target":"C1","id":"main"}},"commits":{"C0":{"parents":[],"id":"C0","rootCommit":true},"C1":{"parents":["C0"],"id":"C1"}},"HEAD":{"target":"C0","id":"HEAD"}}'
       );
+    });
+
+    it("to a commit with --detach option", function () {
+      return expectTreeAsync(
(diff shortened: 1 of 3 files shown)


pcottle:
Will get to this soon! Haven't forgotten about it

pcottle:
Thanks so much for this great contribution @AbdullahButt2611 !! merged and pushing the site now

Issue #1018, opened 2022-10-06
Improve description of clone


timhoffm:
I find this description a bit confusing:

> Technically, git clone in the real world is the command you'll use to create local copies of remote repositories (from github for example). We use this command a bit differently in Learn Git Branching though -- git clone actually makes a remote repository out of your local one. Sure it's technically the opposite meaning of the real command, but it helps build the connection between cloning and remote repository work, so let's just run with it for now.

https://github.com/pcottle/learnGitBranching/blob/a78e4d1327b8cf5eb6bcc2c0300918fab4f26419/src/levels/remote/clone.js#L71

1. I suggest to instead phrase it in one of these ways:

- Don't go into the distinction which one is the remote: "git clone is the command to create a copy of an existing repository"
- Or: invert the language: "Assume that a git repository by someone else exists somewhere (typical on another machine/server). You can use `clone` to get a local copy of that repository."

2. The example uses `git clone` without specifying a remote url. Maybe it's worth to require passing in some dummy url `git clone https://example.com/myrepo.git` to make the example more realistic. Or even make a small example repo on github `git clone https://github.com/pcottle/learnGitBranching-clone-example.git)/, so that the call would work in real life (and users can see the repo in GitHub - though the clone operation in the tutorial is still a dummy of course).

pcottle:
I like your inverted language suggestion. I think technically git doesn't distinguish between which repository is the main source of truth (any repo can send PRs to another repo) but almost all common usage of Git is done with a main source of truth repository.

But yeah for the purposes of the app maybe we just say "you run this command to get a copy" and go from there as you said 😇

timhoffm:
> I like your inverted language suggestion. I think technically git doesn't distinguish between which repository is the main source of truth (any repo can send PRs to another repo) but almost all common usage of Git is done with a main source of truth repository.

There is indeed no main source of truth. However, there is an asymmetry in that the clone has a remote `origin` set up, but not the other way round. Since the concept of a remote is essential, I think I'll try a two-step description.
1. "you run this command to get a copy"
2. "your copy still links to/remembers the original repo as remote `origin`".

I can try and wordsmith this in a PR, but it will be a couple of days before I come to that.

PR #1415, opened 2026-08-22
Opt-in clone implementation

### LINKED ISSUE #1414: Model git clone as remote → local in user-facing lessons
First of all, thank you for building and maintaining Learn Git Branching. I've used it a lot while learning Git, and it's been genuinely helpful — especially for understanding concepts that are much harder to grasp from the command line alone.

While going through the remote lessons again, I noticed that LearnGitBranching currently models `git clone` in the opposite direction from real Git: it creates a remote from the local repository.

I found the historical context in #433, where the local-first lesson structure and data model made this direction easier. #1018 also discusses improving the wording around it. I was wondering if we could go one step further without redesigning the remote subsystem.

My proposal is to keep the existing behavior as a compatibility fallback, but allow the few user-facing lessons where the learner actually types `git clone` to opt into real remote → local semantics. Sandbox behavior and existing internal setup flows could remain unchanged.

I’ve put together a small implementation of this approach and opened a PR directly along with this issue(#1415) so the concrete behavior and tradeoffs are easier to review.

Related: #433, #1018

plt8172:
## Summary

This makes `git clone` behave as remote → local in the user-facing lessons that require the learner to run it, while preserving LGB's existing clone behavior elsewhere.

Fixes #1414
Refs #433, #1018

## Implementation

The affected levels start with a pre-existing `originTree` and an opt-in `clonePending` state.

When `git clone` runs, LGB copies the remote into the local repo, creates the remote's default branch locally, creates `o/<branch>` refs for all remote branches, sets up tracking for the default branch, then reveals the local visualization.

The existing local → remote `makeOrigin()` path remains the fallback when `clonePending` is not set.

## Scope

Only these learner-facing clone flows opt in:

- Clone Intro
- Faking Teamwork
- Diverged History

No broader remote-model refactor, multiple-remotes support, URL/network behavior, or `o/` → `origin/` rename is included.

## Additional fix

The new pre-clone state exposed an existing async ordering issue in `makeOrigin()`: remote-tracking branches could be created before `this.origin` was assigned in the browser.

That setup now runs after `gitEngineReady`.

## Tests

Added coverage for:
- legacy clone compatibility
- remote → local clone behavior
- multi-branch remotes
- non-`main` default branches
- preserving the remote
- reset/re-clone behavior
- async origin initialization

** Existing translations still describe the old reversed clone behavior and can may updated separately.


pcottle:
See #1414 merging now!

PR #1421, opened 2026-09-01
feat: Linux-style Tab autocomplete -- LCP on first press + cycle through matches on repeated Tab

### LINKED ISSUE #842: AutoComplete feature with tab key

bitbad1024:
Builds on #842.

## Background
In #842 there was a good discussion about Tab-completion UX:

- Users typing `git che<Tab>` would land on `git cherry-pick` when they actually
  wanted `git checkout`.
- @Ashark favored a **bash-style UX** — complete up to the longest common
  prefix first, don't "guess" a single match when there are multiple.
- @pcottle said he was **"supportive of something like an additional tab to
  swap over to another suggestion"** (cycle-through).
- Eventually @pcottle closed #842 via commit `095ba99` ("Resolves #842 -- better
  ordering for suggestions"), which re-ordered the suggestion list so common
  commands like `git commit` / `git checkout` rank ahead of `git cherry-pick`.
  That fixed the 80% case, but the cycle-through + LCP UX from the same
  discussion was never implemented.

This PR implements that remaining part, so the ordering hotfix and the UX
improvement stack on top of each other rather than being alternatives.

## What this PR does
Hybrid UX that should feel familiar to users of real shells:

| Situation | Before (after 095ba99 ordering fix) | After |
|---|---|---|
| First Tab press — 1 match only | completes it | completes it (unchanged) |
| First Tab press — N matches + shared LCP longer than input | **picks the 1st match (by order)** | expands input to the **longest common prefix** (bash-style, no guessing) |
| First Tab press — N matches + LCP equals the current input | picks the 1st match by order | enters cycle mode, shows 1st match by order |
| Repeated consecutive Tab presses | **keeps picking the 1st match** (noop after first) | **cycles through all matches** (cycle order = already-improved suggestion order from 095ba99) |
| Any non-Tab keydown between Tabs | — | resets cycle state so the next Tab is a "fresh" first press |
| Semi-separated multi-command input (`cmd1; cmd2<Tab>`) | completes only last token | still only completes last token; prefix before `;` stays intact |

## Implementation
- Small `longestCommonPrefix(strings)` helper.
- 4 new state fields on `CommandPromptView`:
  `_tabMatches`, `_tabIndex`, `_tabLastPrefix`, `_tabLastFullPrefix`.
- Reuses the already-sorted `autoCompleteSuggestionOrder` (so 095ba99's
  ordering improvement feeds directly into the cycle order).
- The existing grey inline hint in `#shadow` is preserved.

## Verification steps
1. Type `git c` then Tab → input expands to the shared LCP `git c` (no guess).
2. Press Tab again → cycles through `git cherry-pick` → `git clone` → `git commit` → …
3. Type `git clo<Tab>` (single match) → completes to `git clone` immediately.
4. Type `git reb<Tab>` → expands to the longest shared prefix; further Tabs cycle.
5. Type `git commit; git ch<Tab>` → only the last command is completed; the
   `git commit; ` prefix is untouched.
6. After cycling, press any letter key (e.g. `o`) then Tab → cycle resets and
   a fresh LCP / match list is computed for the new input.

All 366 Jasmine specs pass:

### DIFF
--- a/src/js/views/commandViews.js
+++ b/src/js/views/commandViews.js
@@ -36,11 +36,35 @@ const allCommandsSorted = autoCompleteSuggestionOrder.concat(
   .filter(command => !!command)
 );
 
+/**
+ * Compute the longest common prefix of an array of strings.
+ * e.g. ['git cherry-pick', 'git clone'] => 'git c'
+ */
+function longestCommonPrefix(strings) {
+  if (!strings || !strings.length) return '';
+  let prefix = strings[0];
+  for (let i = 1; i < strings.length; i++) {
+    let j = 0;
+    while (j < prefix.length && j < strings[i].length && prefix[j] === strings[i][j]) {
+      j++;
+    }
+    prefix = prefix.substring(0, j);
+    if (!prefix) return '';
+  }
+  return prefix;
+}
+
 class CommandPromptView {
   constructor(options) {
     options = options || {};
     this.el = options.el;
     this.$el = $(this.el);
+
+    // Tab-cycle autocomplete state
+    this._tabMatches = null;       // currently matched commands array
+    this._tabIndex = -1;           // current index within the cycle (-1 = not yet cycling)
+    this._tabLastPrefix = '';      // last prefix used to trigger Tab (used to detect reset)
+    this._tabLastFullPrefix = '';  // last full prefix (anything before the final ;) for reassembly
 
     this.initialize();
   }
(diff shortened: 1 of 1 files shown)


pcottle:
Wow this looks great! Awesome work. Merging now :)

pcottle:
(The netlify preview link is quite helpful)

pcottle:
@copilot can you do a small investigation into the rendering of the command line prompt area? it seems to be a bit blurry, like there are two versions of the same text rendered or something, until you move AFTER the `git commit` preview part and start providing arguments. its very subtle. maybe a CSS issue?

PR #1424, opened 2026-09-01
Fix blurry doubled text in the command prompt autocomplete hint

pcottle:
The grey inline suggestion in #shadow contained the *entire* command line (typed characters + suggested remainder) and was painted on top of the real prompt text at 40% opacity, one pixel to the right of it (left: 21px vs p.command's margin-left: 20px). Every character the user had already typed was therefore drawn twice with a 1px horizontal offset, which reads as subtly blurry text -- and it disappears as soon as the input stops matching a command (e.g. once you type arguments after `git commit`), which is exactly the reported symptom.

Two related problems in the same overlay:

- The typed text was inserted with innerHTML while the real prompt escapes it, so anything before a ';' (e.g. `<b>x</b>; git comm`) rendered as markup in the hint, shifting the ghost text out of alignment and putting user-controlled markup into the DOM.
- #shadow did not wrap like p.command (which gets word-wrap: break-word from the generic `p` rule), so a command line long enough to wrap smeared the whole hint across the last visual line and overflowed the bar.

Now the hint renders the typed part in a visibility: hidden span -- it still takes up space so the remainder lands in the right spot, but nothing is painted over the real text -- and the remainder in a visible one, both via textContent. #shadow is aligned and wrapped to match p.command exactly.

Claude-Session: https://claude.ai/code/session_017FgUroE5ouv2N9EkEitzX7

### DIFF
--- a/src/js/views/commandViews.js
+++ b/src/js/views/commandViews.js
@@ -54,6 +54,46 @@ function longestCommonPrefix(strings) {
   return prefix;
 }
 
+/**
+ * The prompt renders spaces as non-breaking so that the fake cursor lines up
+ * with the real characters; the hint has to do the same or the two get out
+ * of sync.
+ */
+function shadowSafeText(text) {
+  return text.replace(/ /g, '\u00a0').replace(/\n/g, '');
+}
+
+/**
+ * Render the grey inline autocomplete hint into #shadow.
+ *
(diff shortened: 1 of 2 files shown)


pcottle:
Screenshot 2026-09-01 at 4 23 47 PM
I wasnt wrong!

PR #1427, opened 2026-09-02
refactor: simplify CommandPromptView key handling and fix 3 bugs

bitbad1024:
Extracted constants:
- KEY_TAB (9), KEY_BACKSPACE (8), KEY_U (85), KEY_W (87)

Bug fixes:
- Fix onKeyUp calling onKeyDown 2-3x per keystroke, causing triple DOM reflow and noticeable keystroke latency. Replaced with direct shadow-hint + prompt sync.
- Fix allCommandsSorted duplicating 'levels' (indexOf > 0 should be === -1).
- Fix Ctrl+W / Alt+Backspace computing lastSpaceIndex from trimmed string but slicing the original, producing wrong indices. Replaced with a two-pointer backward scan.

Structural improvements:
- Extract splitCommands() as single source of truth for ';' parsing.
- Cache shadowEl in initialize instead of per-keystroke querySelector.
- Extract _updateShadowHint(), _handleTab(), _handleDeleteWord(), _setPromptHTML() to flatten nested logic.
- updatePrompt replaces if/else if/else with 3 early returns.
- Classify key once upfront in onKeyDown, dispatch flat.
- onKeyUp keyToFuncMap uses arrow functions instead of .bind(this).

All 366 existing tests pass.

### DIFF
--- a/src/js/views/commandViews.js
+++ b/src/js/views/commandViews.js
@@ -31,10 +31,14 @@ const autoCompleteSuggestionOrder = [
 ];
 
 const allCommandsSorted = autoCompleteSuggestionOrder.concat(
-  // add the rest that aren't in the list above
-  allCommands.map(command => autoCompleteSuggestionOrder.indexOf(command) > 0 ? null : command)
-  .filter(command => !!command)
+  allCommands.filter(command => autoCompleteSuggestionOrder.indexOf(command) === -1)
 );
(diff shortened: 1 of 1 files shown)


pcottle:
Nice work, yeah this logic is getting complex so glad its in a bit better shape now

PR #1433, opened 2026-09-07
Externalize remaining hardcoded UI strings into intl

ranjithrajv:
Most user-facing text already goes through `intl.str()`, but a tail of strings did not — so they could never be translated, and `gulp lintStrings` could not see them. This moves that tail into `strings.js`.

### What changed

- **3 helper-bar tooltips** in `MainHelperBarView.jsx` ("Show commands", "Show available languages", "Follow me on Threads") — the only untranslated text in the main UI chrome.
- **6 hardcoded `msg:` literals** in git and sandbox command handlers (bad numeric argument, `reset solved` warning, alias set/removed, rollup, level-open failure).
- **All 29 `intl.todo()` call sites** converted to `intl.str()`, across `git/commands.js`, `git/index.js`, `mercurial/commands.js`, `sandbox/index.js`, and `sandbox/commands.js`. Where the message interpolated variables, string concatenation became `{param}` interpolation.
- **`intl.todo()` removed**, now that nothing calls it. It was `function(str) { return str; }` — an identity function, so anything routed through it was never translatable and was invisible to the string linter. Deleting it closes the escape hatch: new UI text now either gets a key or shows up as a raw literal in a grep.

38 new keys, `en_US` only, each with a `__desc__` so translators have context. Other locales fall back to `en_US` until someone translates them.

### Notes

- **English output is byte-identical at every call site.** The `{param}` conversions were checked against the same `/\{(.+?)\}/g` regex `intl/index.js` uses.
- Two `msg: ""` blanks (`commandModel.js`, `git/commands.js` in the no-origin `git remote` case) are left alone — intentional empty output, not text.
- `builderViews.js` is untouched. It is level-authoring tooling rather than end-user UI, so it seemed out of scope; happy to include it if you'd rather it were consistent.
- `checkStrings` goes from 112 to 148 good keys, no errors.

### Testing

`npm test` — 366/366 specs pass. `gulp lint` (jshint) and `gulp lintStrings` both clean.

### DIFF
--- a/src/js/git/commands.js
+++ b/src/js/git/commands.js
@@ -51,9 +51,7 @@ var assertNotCheckedOut = function(engine, ref) {
   }
   if (engine.HEAD.get('target') === engine.refs[ref]) {
     throw new GitError({
-      msg: intl.todo(
-        'cannot fetch to ' + ref + ' when checked out on ' + ref
-      )
+      msg: intl.str('git-error-fetch-checked-out', { ref: ref })
     });
   }
 };
@@ -63,9 +61,7 @@ var assertIsBranch = function(engine, ref) {
   var obj = engine.resolveID(ref);
   if (!obj || obj.get('type') !== 'branch') {
     throw new GitError({
-      msg: intl.todo(
-        ref + ' is not a branch'
-      )
+      msg: intl.str('git-error-not-a-branch', { ref: ref })
     });
   }
 };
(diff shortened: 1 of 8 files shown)


pcottle:
Thanks a ton for this! You can see that these "todo"s were never addressed even a decade later 😅 this will set up the app for a better i18n experience for sure

PR #1436, opened 2026-09-09
Fix undo/gitCommandsIssued desync in Level (#1298)

Jeevanmerkaji:
## What's going on

`Level.afterCommandCB()` pushes to `undoStack`/`gitCommandsIssued` for any command that "counts" by name (`doesCommandCountTowardsTotal`), even if the command left the tree completely unchanged — e.g. a cancelled or no-op `git rebase -i` reorder. `Level.undo()` separately pops `gitCommandsIssued` **unconditionally**, even when `undoStack` is already empty and the base `Sandbox.undo()` has nothing left to restore.

Either path lets the two arrays drift out of lockstep with each other and with the real number of undoable tree states. That matches what's described in #1298 on the "Juggling Commits" level (which revolves around `git rebase -i`): commits disappearing after a few undo/redo cycles, and further undos not bringing them back.

## The fix

Pulled the lockstep bookkeeping into its own module, `level/undoBookkeeping.js`:
- `recordCommand()` only pushes an undo entry if the command both counts *and* actually changed the tree.
- `shouldPopIssuedCommand()` tells `Level.undo()` to only pop `gitCommandsIssued` when `undoStack` genuinely has something for the base class to restore.

`level/index.js` now delegates to these instead of doing the bookkeeping inline.

## Why a new file instead of editing in place

`level/index.js` requires a `.jsx` React view (`LevelToolbarView`), which the plain jasmine test runner can't transpile — so nothing that imports that file has ever been directly unit tested (`levels.spec.js` even explicitly skips any level whose solution uses `git rebase -i`, for the same underlying reason). This follows the same pattern already used for `levelRegexMap.js` (see the comment above it) to keep pure logic testable without pulling in the view layer.

## Testing

New `__tests__/levelUndoBookkeeping.spec.js` exercises the real logic `Level` now delegates to, including the exact "a command that counts but no-ops, then undo" sequence from the bug report.

Verified red→green: the two bug-specific tests fail against the old inline logic and pass against the fix. Confirmed no regressions: `yarn build` (jasmine + jshint + string lint) is fully green — all 366 prior specs pass unmodified, plus the 4 new ones.

```
Executed 370 of 370 specs SUCCESS
```

### DIFF
--- a/__tests__/levelUndoBookkeeping.spec.js
+++ b/__tests__/levelUndoBookkeeping.spec.js
@@ -1,0 +1,107 @@
+var undoBookkeeping = require('../src/js/level/undoBookkeeping');
+
+// Regression test for issue #1298 ("Undo Bug in Juggling Commits #1"): after
+// a few undo/redo cycles, commits went missing and further undos couldn't
+// bring them back.
+//
+// Level (src/js/level/index.js) keeps two arrays in lockstep: undoStack
+// (tree snapshots to restore) and gitCommandsIssued (command strings, used
+// for the git-golf command count and level-completion tracking). The actual
+// bug lived in that class's undo()/afterCommandCB(), but that file also
+// requires a .jsx React view, which this test runner can't transpile -- so
+// the lockstep logic was pulled into its own plain module
+// (level/undoBookkeeping.js) that Level now delegates to. This tests that
+// module directly.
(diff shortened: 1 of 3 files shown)


pcottle:
Great catch and thanks for the thorough explanation/testing! Sorry the app is half in react (and hence some tsx files are littered around :P )

PR #1437, opened 2026-09-09
Add Italian translations for git restore hints

SimoDP:


pcottle:
Thank you so much! Grazie!!

PR #1444, opened 2026-09-11
Add a shareable completion certificate

pcottle:
## Summary

- show a dynamically rendered completion certificate after all levels are solved
- let learners add their name, copy the certificate image, or download it as a PNG
- summarize six concepts covered by the curriculum without crowding the certificate
- keep the existing X, LinkedIn, and Facebook sharing options
- center the recipient name and commit graph, and provide an explicit Close button
- remember whether the automatic celebration has already been shown
- expose a `certificate` preview command only when the app is opened from a local `file://` URL

Closes discussion https://github.com/pcottle/learnGitBranching/discussions/1443

## Testing

- `npm test` (376 specs)
- `npm run build` (production bundle, tests, lint, and string validation)

### DIFF
--- a/__tests__/LevelStore.spec.js
+++ b/__tests__/LevelStore.spec.js
@@ -93,6 +93,28 @@ describe('this store', function() {
 
     LevelActions.resetLevelsSolved();
   });
+
+  it('reports completion only after every level is solved', function() {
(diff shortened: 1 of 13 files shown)


pcottle:
learn-git-branching-certificate-peter-cottle (3)

PR #1456, opened 2026-09-18
perf: cut initial JS ~77% by lazy-loading levels, locales and vendor code

ranjithrajv:
## Summary

Cuts the initial JavaScript payload by **~77% (≈1.2 MB gzip → ≈276 KB gzip)** by code-splitting level definitions, locale strings, and third-party code so each is fetched only when needed. Also resolves the active locale (and remembers it) before first paint, adds a shared loading overlay, and prefetches levels on intent.

## What changed

- **Lazy levels.** `src/levels/index.js` exposes a declarative `levelFiles` map. A generated manifest (`scripts/generateLevelManifest.js`) carries only `id`/`name` for the level map; `LevelStore.loadLevel(id)` dynamically imports the full definition when a level is opened. `startLevel` and the builder's edit path await it.
- **Lazy locales.** `scripts/generateIntlStrings.js` splits `intl/strings.js` into per-locale modules. `intl` synchronously loads `en_US` as a fallback and lazily imports the active locale; `bootstrap()` awaits it before `init()` so there is no English flash.
- **Remembered locale.** Persisted in `localStorage` as `{loc, at}` (180-day TTL, legacy plain-string format accepted). Resolution order: `?locale=` → stored → browser header.
- **Loading UX.** One reusable `#appLoading` overlay for slow startup and on-demand level loads; 250 ms delayed fade for startup (fast loads never flash it), immediate for level opens, with a 15 s safety timeout.
- **Prefetch on intent.** The next level after solving, or a level hovered in the map.
- **Vendor chunk.** Third-party code in a stable `vendor` chunk, cacheable across app-only deploys.

Generated artifacts (`src/levels/generated/`, `src/js/intl/generated/`) are gitignored and produced by `yarn generate`, wired into `predev`/`prebuild`/`pretest`.

## Impact

| | before | after |
|---|---|---|
| Initial JS (gzip) | ~1,201 KB | **~276 KB** |
| First load incl. CSS/runtime/locale | ~1,210 KB | **~285 KB** |

Repeat visitors re-download ~130 KB gzip on app-only deploys instead of ~276 KB. Total bytes over a full session are unchanged — content is deferred, not removed.

## Testing

- `yarn test`: **407 specs, 0 failures**; `yarn lint`: 0 errors.
- **Every commit in the stack passes tests + lint in isolation** (verified in throwaway worktrees), so it is bisectable.
- Browser smoke: sandbox loads, levels open and solve on demand, `fr_FR`/`de_DE` locale chunks load, and locale selection persists across reloads.

## Notes / tradeoffs

- Non-English first paint waits one RTT for the locale chunk; live language switches still flash the previous language briefly.
- A stored locale overrides `navigator.language` until it expires (180 days) or the user switches again.
- No new runtime dependencies; the build gained two code generators and a vendor chunk.


pcottle:
This is very exciting! i think the tradeoff on local switching is somewhat acceptable -- hopefully most people just land on the locale they want via the browser headers.

ill take a look at everything else

src/js/app/index.js
```diff
@@ -314,9 +309,44 @@ function changeLocaleFromHeaders(langString) {
...
+    try {
+      initFn();
+    } finally {
+      clearTimeout(safety);
+      appLoading.hideAppLoading();
+    }
+  });
+}
+
+// Resolve the active locale before first render: an explicit ?locale= wins,
```
pcottle, line 328: glad we kept this behavior

src/js/stores/LevelStore.js
```diff
@@ -3,8 +3,9 @@
 var AppConstants = require('../constants/AppConstants');
 var AppDispatcher = require('../dispatcher/AppDispatcher');
 var EventEmitter = require('events').EventEmitter;
-var levelSequences = require('../../levels').levelSequences;
-var sequenceInfo = require('../../levels').sequenceInfo;
+var manifest = require('../../levels/generated/manifest');
```
pcottle, line 6: and nowadays the level manifest changes very very rarely

pcottle:
ooooh one thing i did notice on the preview (https://deploy-preview-1456--xenodochial-hugle-b9ec84.netlify.app/?locale=es_AR) is that when i loaded up a level, the whole app flashes blank and shows a loading indicator, and THEN the level pops in. the latency is totally fine but the visual flash is undesirable -- i might try to tweak that!

ranjithrajv:
> the latency is totally fine but the visual flash is undesirable -- i might try to tweak that!

agree. makes sense.

pcottle:
also brought back a bit of the locale fallback logic since apparently it still has an impact

```

• We still benefit from fallback, although less than the aggregate percentages suggest. The runtime fallback applies only to the 176 UI strings—not the level translations.

  Exact impact:

  - zh_TW → zh_CN: recovers 12 of 70 missing UI strings, including Git status messages, progress sharing, close-window, and back-button copy.
  - es_MX → es_ES: recovers 2 of 72 missing strings—the main and remote level-tab labels.
  - es_AR → es_ES: recovers 0 of 58, so this fallback is currently redundant.

  The application remains functional without these because English is the final fallback. So the question is compatibility and language quality, not correctness.
  ```

pcottle:
ah dang I noticed this broke the dev workflow somehow -- levels wont load anymore (I think due to the inheritance we have). debugging...

pcottle:
uh oh, I merged this and it broke prod somehow. let me revert for now and double check...

pcottle:
ok I think I got this out in prod working. I had to handle vite preload errors since we didnt have push safety across deploys. but it should work now!

pcottle:
Let me know once you take it for a spin if it works on your end! You should see the smaller payload, per level fetching, and also some preloading of level n+1 when level n is loaded

PR #1458, opened 2026-09-20
a11y: dialogs, landmarks, live regions, contrast, keyboard rebase and graph summary

ranjithrajv:
## Summary

Accessibility pass addressing the open checklist in #960, plus the surrounding gaps. No behavior/feature changes for sighted mouse users beyond slightly darker button/active colors and a visible focus ring.

## What changed

**Dialogs & focus**
- `role="dialog"` + `aria-modal="true"` on modals, with accessible names taken from their titles.
- Focus moves into a dialog on open and returns to the previous element on close; hidden dialogs are `aria-hidden` + `inert`, so they leave the accessibility tree.
- Window close controls are `role="button"`, focusable, and operable with Enter/Space.

**Landmarks & live regions**
- `role="main"` on the app shell, `<nav>` for the helper bar, labelled `region` for the terminal.
- Terminal output is a `role="log" aria-live="polite"` region, so command results are announced.

**Commit graph**
- The SVG is decorative (`aria-hidden`); a plain-text summary of HEAD/branches/tags/commits is published to a labelled region (`#a11yGraphSummary`) on every render and on engine-ready.

**Keyboard**
- Interactive-rebase entries gain **Move up / Move down** buttons and a pick/omit toggle with `aria-pressed`; order is model-authored and synced with jquery-ui drags.
- Global two-tone `:focus-visible` ring; `prefers-reduced-motion` handling for CSS motion.

**Contrast**
- Darker gradient and `:active` states for AA; larger helper text; omitted rebase rows are dimmed via a line-through + reduced opacity on the text only, keeping the controls at full contrast.

**i18n & tooling**
- New `a11y-*` strings; the graph summary is routed through `intl.str`.
- Fixed `checkStrings` key regex to accept digits (the `a11y-*` keys were falsely reported as unused).

**Models/tests**
- Extracted `RebaseEntry`/`RebaseEntryCollection` into a DOM-free module with specs (reorder, drag-order sync, clamping, pick/omit ordering).

## Verification
- `yarn test`: **384 specs, 0 failures**; `yarn lint`: 0 errors; `yarn lint:strings`: clean.
- Lighthouse accessibility: **100** on sandbox, level dialog, `help` alert, and level list.
- Fresh-load graph summary populates without any command.

## Notes / remaining
- The graph summary region is intentionally **not** a live region, to avoid double announcements with the terminal log.
- JS-driven graph animations are still not reduced under `prefers-reduced-motion` (CSS motion is); that's a WCAG AAA follow-up.
- New labels/strings ship `en_US` only (fall back to English); translations welcome.
- Real screen-reader passes (VoiceOver/NVDA) are still recommended on top of the automated audit.

Addresses #960.


src/js/intl/checkStrings.js
```diff
@@ -6,7 +6,7 @@ var readDirDeep = require('../util/readDirDeep').readDirDeep;
 var { strings } = require('../intl/strings');
 
 var canonicalLocale = 'en_US';
-var easyRegex = /intl\.str\(\s*'([a-zA-Z\-]+)'/g;
+var easyRegex = /intl\.str\(\s*'([a-zA-Z0-9\-]+)'/g;
```
pcottle, line 9: oh do we have intl strings with numbers in them? my bad. ah the `a11y` ones i see

src/js/util/a11y.js
```diff
@@ -0,0 +1,88 @@
+"use strict";
+
+var intl = require('../intl');
+
+/**
+ * Accessibility helpers for the commit graph.
+ *
+ * The graph is an SVG, which screen readers cannot meaningfully describe, so we
+ * publish a plain-text summary of the current tree into a labelled region that
```
pcottle, line 9: nice! maybe we could have some debug output into the console or something to see what this shows

src/js/util/constants.js
```diff
@@ -12,13 +12,19 @@ var VIEWPORT = {
   minHeight: 600
 };
 
+// Users who ask for reduced motion get near-instant graph updates instead of
+// the full animation. Read once at load; a mid-session change needs a reload.
```
pcottle, line 16: mid session change needing a reload is fine imo

src/js/views/index.js
```diff
@@ -227,6 +227,14 @@ class ModalView {
...
     this.el = document.createElement(this.tagName);
     this.el.className = this.className;
+    this.el.setAttribute('role', 'dialog');
+    this.el.setAttribute('aria-modal', 'true');
+    // hidden until shown; keeps it out of the accessibility tree
+    this.el.setAttribute('aria-hidden', 'true');
+    this.el.setAttribute('inert', '');
+    this.el.setAttribute('tabindex', '-1');
+    // dialogs need an accessible name; callers pass one via options.label
+    this.el.setAttribute('aria-label', options.label || intl.str('a11y-dialog'));
```
pcottle, line 237: could these be set in the html template rather than at runtime here? everything but the a11y string

src/js/views/index.js
```diff
@@ -294,12 +312,16 @@ class ModalView {
         if (child.classList.contains('modalView')) return;
         if (!child.hasAttribute('inert')) child.setAttribute('inert', '');
       });
+      this.el.removeAttribute('aria-hidden');
```
pcottle, line 315: oh i see we might be managing them, nm

src/js/visuals/index.js
```diff
@@ -435,12 +436,18 @@ GitVisuals.prototype.refreshTree = function(speed) {
 
   // this method can only be called after graphics are rendered
   this.fullCalc();
+  if (!this.gitEngine.isOrigin()) {
```
pcottle, line 439: we might need the same behavior on origin trees for the description, but im not 100% sure

src/js/util/a11y.js
pcottle:
```
document.getElementById('a11yGraphSummary').textContent
'HEAD is on branch main. Branches: main at C3. Commits: C0 (root); C1 after C0; C2 after C1; C3 after C2.'
```

cool!

pcottle:
Two small requests if thats ok:

1) add some tests for describe tree so it doesnt regress
2) right now the modal logic removes `inert` from every body child, which might not be safe (if some other child had inert for a specific reason). worth revisiting?

